PR #2, opened 2017-06-06
match link colors with alert color

tato123:
Update to `Alert` to add alert colors to child links

This change revs the version to `0.6.4-alpha.4`

### DIFF
--- a/src/components/Alert.js
+++ b/src/components/Alert.js
@@ -10,6 +10,32 @@ const Alert = styled.div`
   font-weight: 500;
   width: auto;
   line-height: 1em;
+  
+  a {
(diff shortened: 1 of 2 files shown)


src/components/Alert.js
```diff
@@ -10,6 +10,32 @@ const Alert = styled.div`
   font-weight: 500;
   width: auto;
   line-height: 1em;
+  
+  a {
+    color: ${({ theme, type }) => {
```
a-type: Could we possibly just use `inherit` for this?

tato123: ... wow, completely overlooked that, funny enough I used inherit in my testing, you're right

PR #4, opened 2017-06-08
V1 beta

a-type:
Massively reorganizes the repo to be easier to consume.

Documentation is still upcoming.

Wanted to go ahead and open this PR so everyone can see where we're going.

A high level overview of the changes:

* Components are now organized in a flat manner, mirroring the component names described in the Design System. Related components are now grouped as sub-components. This includes some pretty high-profile renames, like `Link` -> `Anchor`!
* All components, extensions, themes are now exported from the root index. A script maintains this index file. If you know the name of a component, you can just import it directly from the library like `import { ButtonField } from '@bandwidth/shared-components';`
* The library now ships with Overpass font and a file you can import called `bootstrap` which will load in global CSS.
* New component: `Navigation`-- an overhauled version of `TabGroup` and `TopNav`, this creates the nice headers we see in our design mockups.

Note that some component renames are tricky. For instance, `InlineCode` is now `Code`, and `Code` is now `CodeBlock`. I'll try to detail the migration path soon in documentation.

I don't expect anyone to read through all these file moves. You can go ahead an install the beta by running `npm i @bandwidth/shared-components@beta` and test it out yourself. Also, wait for documentation, which will hopefully clarify everything.


src/components/Navigation/LogoHeader.js
```diff
@@ -10,15 +10,15 @@ const Container = styled.div`
 `;
 
 const Text = styled.span`
-  font-size: ${({ theme }) => theme.header.fontSize};
-  font-weight: ${({ theme }) => theme.header.fontWeight};
+  font-size: 22px;
```
tato123: This change is a little concerning, moving font information locally to the component will make it more challenging in the future to support responsive designs such as mobile.

src/components/Navigation/NavigationItems.js
```diff
@@ -0,0 +1,72 @@
...
+  to: PropTypes.string,
+  onClick: PropTypes.func,
+  content: PropTypes.node.isRequired,
+  exact: PropTypes.bool,
+}));
+
+const NavigationItems = ({ links }) => (
+  <Container>
+    {links.map((link) => (
+      <Anchor key={`${link.content}_${link.to}`} to={link.to} onClick={link.onClick} exact={link.exact}>
```
tato123: Since the `link.to` property is not required it seems like it would be possible to have duplicate keys if you receive multiple links with the same content value and no to value (could be useful if you wanted to create a set of navigation item links that function as callbacks that don't have an actual destination).

Same situation with multiple links with the same to value (but that might be a little odd, can't think of a reason to do that intentionally).

src/components/SidebarList/SidebarList.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    return null;
+  };
+
+  renderItems = () => {
+    const { selectedIndex } = this.props;
+
+    return React.Children.map(
+      this.props.children.filter((child) => child !== null),
```
tato123: `this.props.children.filter` relies on type array however if there is only a single child element then `this.props.children` is set to the react element itself / a factory function. This will throw an error on an undefined method.

tools/generateIndices.js
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * Please see /tools/generateIndices.js to regenerate after structural changes or component additions / deletions
+ */
+
+`];
+const exportItems = [];
+
+const src = path.resolve(__dirname, '..', 'src');
+
+const appendExports = (dir, subDirs, files) => {
```
tato123: FYI [Future]: This should likely be converted to a webpack plugin since this is the sort of thing it handles, then you would receive a stream of files that you could generate the indices off of. It would allow this solution to be a little more flexible.

src/components/Navigation/LogoHeader.js
a-type: In my experiments on side projects, I've found that it's best for the theme to only have marketing-level concerns - colors, fonts, etc.

If I want to change LogoHeader's font size, my first thought is to go to `LogoHeader.js`. Having to then backtrack to someplace in a `theme` file is a maintenance headache which is the result of (imo) an over-reaching abstraction from the theme on my part.

If we were to make this component responsive, we would just add a media query in this block. The media breakpoint could certainly be provided by some external constant to keep things consistent across the application, but why should any file but `LogoHeader.js` be concerned with how the LogoHeader reacts to different media sizes? That's my thinking at this point.

If we needed to provide a more top-level responsiveness pattern, I'd still recommend creating a new HOC-based system to provide those details to components rather than packing things into the theme.

src/components/Navigation/NavigationItems.js
a-type: It's true that you could introduce identical keys, but I'm not really seeing the use case. I don't think a design would ever call for identical navigation tabs within the same row.

src/components/SidebarList/SidebarList.js
a-type: Ah. Good call.

tools/generateIndices.js
a-type: Webpack isn't involved in this library anymore, it's all just Babel now. I wouldn't be against re-introducing it, just need to figure out exactly how to do that.

The original reason I moved to Babel was to keep 1:1 source:transpiled parity, since the primary consumption use case was to delve into the subdirectories to find exact files. But if we move to consuming components directly from the library root, I guess we could switch to webpack for distribution builds. Worth looking into again.

PR #8, opened 2017-06-22
A few changes to prepare for the demo workshop. Root component.

a-type:
The Root component makes it easier to get started with the library. It boostraps the global CSS and provides the default theme.

Other than that, just added a few tweaks. `FlexFields` now has a default value so it doesn't complain if you use it alone.


src/components/Root/Root.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { ThemeProvider } from 'styled-components';
+import theme from '../../theme';
+import '../../bootstrap'; // sets base CSS
+
+const Root = ({ children }) => (
```
tato123: Choosing to use `Root` for the component name seems a little inconsistent with both `styled-components` and other component libraries like material-ui react. Since this is defaulting the theme provider I would suggest something like `BandwidthThemeProvider`. This would be similar to the root node found in the [react material ui component library](http://www.material-ui.com/#/get-started/usage) `MuiThemeProvider`

a-type: Good call. Can definitely do that.

PR #9, opened 2017-06-26
Support id and className props on all components

a-type:
Adds propTypes and prop support for `id` and `className` to all components, as well as propType documentation for most propTypes.


a-type:
Including all for review mostly for visibility.

PR #10, opened 2017-06-27
update alerts to wrap to the right of icon

tato123:
Wraps the text content of the alert to the right of the icon.

### DIFF
--- a/src/components/Alert/Alert.js
+++ b/src/components/Alert/Alert.js
@@ -11,6 +11,7 @@ const AlertContainer = styled.div`
   font-weight: 500;
   width: auto;
   line-height: 1em;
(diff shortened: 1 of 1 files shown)


src/components/Alert/Alert.js
```diff
@@ -111,13 +118,20 @@ Alert.propTypes = {
   id: PropTypes.string,
 };
 
-Alert.defaultProps = {
+AlertContainer.defaultProps = {
```
a-type: Should these props also be exposed on `Alert` now?

tato123: yup, you're right, completely overlooked that

PR #12, opened 2017-06-27
Make Anchor work with external urls

a-type:
Updates styled-components to take advantage of the ability to swap out the backing component. Now Anchor will render a regular <a> if it detects an absolute link.

### DIFF
--- a/src/components/Anchor/Anchor.js
+++ b/src/components/Anchor/Anchor.js
@@ -77,7 +77,7 @@ const inferType = (children) => {
   return 'wrap';
 };
 
-const isExternal = (to) => /^(https?)*:\/\//.test(to);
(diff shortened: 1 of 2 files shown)


src/components/Anchor/Anchor.js
```diff
@@ -71,7 +75,9 @@ const inferType = (children) => {
   }
 
   return 'wrap';
-}
+};
+
+const isExternal = (to) => /^(https?)*:\/\//.test(to);
```
tato123: It seems like this regex is just testing that a uri has `https`, what about `http` and `///_url_goes_here` links?

a-type: It matches `https://foo` and `http://foo` (since the `?` makes the `s` optional). I just updated it to also match `//foo` for protocol-irrelevant links.

PR #14, opened 2017-07-10
fix-money check null money value

lunxiao:
when value is null, return null instead of calling function on value


lunxiao:
I added the check only because it failed when I used the Money component initially without checking. I guess "required" does not guarantee it is not null.

a-type:
Yeah, it's not a guarantee, but if a prop is required and you use the component without providing it, that's a problem with your usage, not something the component needs to fix.

The default seems like the right solution. Now if you just do `<Money />` it will be $0, which makes sense.

PR #15, opened 2017-07-13
Flow (v1.1.0)

a-type:
Yet another massive changeset.

This introduces the Flow system.

Flow
====

      
## The Flow System

Flow is a layout system which is intended to capture the rules which have surfaced in various UX mockups into tested, repeatable code. With Flow, each UI element is annotated with an optional label and help text component. Vertical spacing is kept consistent on every row. Horizontal alignment has sane defaults and a lot of options for fine-tuning. Flow is primarily designed for use inside forms to layout inputs, but it can be used anywhere you need it.

### Recognizing a Flow.Item

Flow Items have a distinctive layout and appearance. When you see one in a design mockup, use the Flow system.

A Flow Item

Flow Items have a label, help text, and content which has a fixed height. Sometimes they might be missing a label, or help text, or both. As a rule of thumb, though, anything which is on the same horizontal level as another Flow Item should also be a Flow Item.

#### Simple Row

Simple Recipe

```
<Flow>
  <Flow.Item label="User ID:" helpText={userIdHelpText}>
    <Input value={userId} disabled />
  </Flow.Item>
  <Flow.Item label="API Token and Secret:" helpText={apiHelpText}>
    <Button>Show Token and Secret</Button>
  </Flow.Item>
</Flow>
```

#### Precise Field Alignment

Sometimes you need to use nesting to achieve a precise alignment of various rows of fields. Particularly, if you want to break one column down into two columns on a certain row, you need to use nesting to line them up prefectly.

Alignment Recipe

(Item contents omitted for brevity)

```
<Flow>
  <Flow.Row>
    <Flow.Item label="Company" />
    <Flow.Item label="Address" />
  </Flow.Row>
  <Flow.Row>
    <Flow.Item label="City" />
    <Flow.Row>
      <Flow.Item label="State" />
      <Flow.Item label="Zip" />
    </Flow.Row>
  </Flow.Row>
</Flow>
```

#### Redux Form

Flow includes built-in Redux Form fields for all our Input types. Reference them from `Flow.fields`. These components will take care of all transformations necessary for Redux Form Field usage.

Form Recipe

```
// in a redux-form HOC
<Form onSubmit={this.props.handleSubmit}>
  <Flow>
    <Flow.Row>
      <Field
        name="name"
        label="Application name:"
        required
        component={Flow.fields.Input}
      />
      <Field
        name="id"
        label="Application ID:"
        disabled
        placeholder="We create this for you"
        component={Flow.fields.Input}
      />
    </Flow.Row>
    <Flow.Row>
      <Field
        name="callbackRequestMethod"
        label="Callback request method:"
        required
        options={['POST', 'GET']}
        component={Flow.fields.RadioGroup}
      />
      <Field
        name="type"
        label="Application type:"
        required
        options={['MESSAGING', 'VOICE', 'BOTH']}
        component={Flow.fields.RadioGroup}
      />
    </Flow.Row>
  </Flow>
</Form>
```

#### Form Controls

Usually in a form, submit buttons and other controls lay out differently. There are a variety of possibilities, but all are possible with Flow! Here are some examples:

Left Recipe

```
<Flow>
  <Flow.Row alignment="left">
    <Flow.Item>
      <Button>Update</Button>
    </Flow.Item>
    <Flow.Item>
      <Anchor>Cancel</Anchor>
    </Flow.Item>
  </Flow.Row>
</Flow>
```

Spaced Recipe

```
<Flow>
  <Flow.Row>
    <Flow.Item alignment="left">
      <Anchor>Show optional fields</Anchor>
    </Flow.Item>
    <Flow.Row alignment="right">
      <Flow.Item>
        <Anchor>Cancel</Anchor>
      </Flow.Item>
      <Flow.Item>
        <Button disabled>Create</Button>
      </Flow.Item>
    </Flow.Row>
  </Flow.Row>
</Flow>
```


a-type:
To reviewers: no, I don't necessarily think you need to read all 6k lines. Most of it is docs. Please be sure to read `Flow.js`, `FlowRow.js`, `FlowItem.js`, and `Flow/fields/flowField.js`. These are the core of the code. Also check out changes to `Select.js`. There are new unit tests for components in this PR!

src/layouts/Form/FieldWrapper.js
```diff
@@ -68,6 +69,10 @@ class FieldWrapper extends React.Component {
     error: null,
   };
 
+  componentDidMount() {
+    warning(true, 'DEPRECATED: the Field system is now deprecated. Please use Flow instead. (from FieldWrapper)');
```
tato123: FYI: May want to consider using an HOC, this would allow you to deprecate a component either within the component or prior to being exported by the library, this would allow you to do things like build a helper to deprecate components (i.e. you could have a helper that reads from read from a JSON file and automatically deprecates components prior to export)

a-type: That's a nice idea. I think this is probably good enough for this round, but next time things get deprecated I'll definitely look into it.

docs/components/FlowItem.md
```diff
@@ -0,0 +1,58 @@
+FlowItem
```
lunxiao: what's the difference between FlowInput and FlowItem?

a-type: `FlowItem` is a wrapper for any generic Component that provides a label, help text, and layout properties. `FlowInput` (which should be used by referencing `Flow.fields.Input`) is a prebuilt `FlowItem` with an `Input` inside.

docs/components/FlowInput.md
```diff
@@ -0,0 +1,4 @@
+FlowInput
```
lunxiao: The samples show usage of <Button> <Fields> under <FlowItem>. When and where will FlowInput be used?

a-type: Oh, yeah, I forgot the docs on these.

PR #16, opened 2017-07-24
Table improvements

a-type:
* Makes several improvements to the current Table components to bring them more in line with design system guidelines. Includes adding scroll shadows, a better loading animation, better header link styling, correct ordering of sort arrows.
* Adds a new component, SimpleTable (available as Table.Simple) which accepts data and a few processing functions and creates a full table with some basic control abilities.


docs/components/TableRow.md
```diff
@@ -8,8 +8,8 @@ Props
 Prop                  | Type     | Default                   | Required | Description
 --------------------- | -------- | ------------------------- | -------- | -----------
 children|node||Yes|Contents of the row - should be cells.
-loading|bool|false|No|[PENDING DEPRECATION] renders the row as semitransparent.
 className|string|null|No|Adds a class name to the element.
 id|string|null|No|Adds an id to the element.
+onClick|func|() => null|No|Handler that is called whenever the row is clicked.
```
lunxiao: Does this mean that user can click anywhere on this row to invoke this handler?

a-type: Yeah, pretty much. Won't work well with links present within the row, but that's not the use case this table is meant to target.

lunxiao: That is nice. There might be a use case for this feature in Q2C.

a-type:
Ok, I've tested this with Catapult and it's non-breaking. Functionality seems to be good. I'll go ahead and release.

PR #18, opened 2017-08-25
Update SimpleTable.js

avbel:
Fixed warning 
`Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of 
 SimpleTable`

image


a-type:
Released in `v1.1.15`

PR #19, opened 2017-09-06
Updates shared components to use style-guidist for documentation

tato123:
The majority of the file changes are removal of the old documentation in favor of documentation generated by [react-styleguidist](https://github.com/styleguidist/react-styleguidist).

- PR should not modify the behavior or logic of any existing components
- PR adds markdown files in the container / component folders which is used by style-guidist auto-generated process
- All debugging tools (xray) are only rendered within the style-guidist documentation and are not exported to libraries

**Important things to note**
- This PR does not match 1:1 with the existing documentation. This is because not all `usage` statements have been transferred to corresponding markdown documents 
- Files that exported only a styled-component have been wrapped by an arrow function. This is due to a limitation in the way react-prop docs and style-guidist docs discovers components. This should not impact functionality since the same component is exported.


a-type:
Wrapping the styled-components was on my todo list anyway. Eventually I want to wrap all components in a HOC so we can have more fine-grained control over behaviors across the app.

However, I'd warn that this can cause bugs in applications which use the library, and internally in the library itself. `styled-components` allows you to utilize a generated component as a classname within another component definition, ala:

```
const Foo = styled.div``;
const Bar = styled.div`
  & > ${Foo} { color: red; }
`;
```

If anyone is using our components this way (and I'm pretty sure we are somewhere), that will break. For that reason I'd say we make this a major release. I'll pull down this version and test it with Catapult as well to see if anything shows up.

tato123:
This is good to know, I had no idea that this would cause breakage. As a result I've updated to a major version change based on your feedback. Additionally I have changed all components that originally exported a `styled` component to now return a `<component>.Styled` property as suggested.

PR #21, opened 2017-09-08
KAT-1881 keep table row exapnded after contents changed by actions pe…

lunxiao:
…rformed on the row

### DIFF
--- a/src/components/Table/SimpleTable.js
+++ b/src/components/Table/SimpleTable.js
@@ -44,6 +44,10 @@ class SimpleTable extends React.Component {
      * Indicates whether the table should render a loading state.
      */
     loading: PropTypes.bool,
+    /**
+    * A function that provides an unique id for a row item.
+    */
+    getId: PropTypes.func,
   };
 
   static defaultProps = {
@@ -55,12 +59,23 @@ class SimpleTable extends React.Component {
     renderDetails: null,
     columns: null,
     onSortChanged: () => null,
+    getId: undefined,
     loading: false,
   };
 
   state = {
     sortOrders: {},
-    detailsItem: null,
+    detailsItemId: null,
+  };
+
+  getId = (item) => {
+    if (_.isString(item)) {
+      return item;
+    }
+    if (_.isFunction(item.get)) {
+      return item.get('id');
+    }
+    return item.id || JSON.stringify(item);
   };
 
   blankSortOrders = () => this.props.columns.reduce(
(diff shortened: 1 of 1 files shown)


src/components/Table/SimpleTable.js
```diff
@@ -55,12 +59,23 @@ class SimpleTable extends React.Component {
...
     loading: false,
   };
 
   state = {
     sortOrders: {},
-    detailsItem: null,
+    detailsItemId: null,
+  };
+
+  getId = (item) => {
```
a-type: You can just provide this function as the `defaultProps` value for `getId` and skip the try / fallback logic.

src/components/Table/SimpleTable.js
```diff
@@ -125,13 +143,15 @@ class SimpleTable extends React.Component {
 
   renderRowDetails = () => {
     const itemIdx = this.indexOfDetailsItem();
-    if (!this.props.renderDetails || !this.state.detailsItem || itemIdx === -1) {
+    const item = this.props.items[itemIdx];
```
a-type: Move this line below the conditional check; we don't need `item` until we're sure we will render it.

src/components/Table/SimpleTable.js
```diff
@@ -55,12 +59,21 @@ class SimpleTable extends React.Component {
     renderDetails: null,
     columns: null,
     onSortChanged: () => null,
+    getId: (item) => {
```
tato123: This is concerning as a function in the shared-components library. Shared components should not assume the shape of `item`. I think this should be a `noop` function  ( `getId: () => null`) instead. This would place the responsibility of identifying the id on the calling application.

a-type: I suggested it to provide some sane out-of-the-box behavior since this component is focused on boostrapping a table from some data without a lot of effort. Without some default implementation, you wouldn't be able to expand details of a row at all. I don't really think it's problematic to take a guess at something useful and fallback to comparing the strict values of the items with stringify.

tato123: Ok, I'm willing to concede that some sort of default may seem helpful from a usage perspective. I'm still a little concerned about this within the shared-components library but not enough to reject it.

a-type: I think this component is a bit of an oddball. We could perhaps spin off a scl-recipes library that combines components into common bootstrapped patterns maybe.

PR #22, opened 2017-09-08
Wrap everything in a HOC

a-type:
Cause it's the new hipster trend.

But really, it powers a more efficient and compatible way to do modifications like "small", "large", perhaps even "secondary" down the line.

This PR introduces the 'mod' system and two provider components, <Small> and <Large>. It also modifies a few components to interact with these providers and change their sizes appropriately.

The idea here is that we can create provider components which modify blocks of the UI. It also avoids having to repeat the same props across various components which implement certain patterns, like changing size. Instead of adding propTypes for `size` to Button, Accordion, Table, etc etc, we can just assume that these will be provided via `mods` and write the logic without worrying about the plumbing or documentation.

It's proven slightly clunky in initial implementations, I admit, since you still have to pass down mods to child components defined within the component itself. I'm open for discussion about whether this is the optimal approach to handling this feature.

Otherwise, a few bugfixes here and there, and a lot of documentation work for some components that got overlooked.


src/components/Accordion/Accordion.js
```diff
@@ -10,19 +11,26 @@ export const Container = styled.div`
...
-  padding: ${({ theme }) => theme.accordion.padding};
-  color: ${({ theme }) => theme.accordion.labelFG};
-  font-family: ${({ theme }) => theme.accordion.labelFontFamily};
-  font-size: ${({ theme }) => theme.accordion.labelFontSize};
-  text-transform: ${({ theme }) => theme.accordion.textTransform};
-  font-weight: ${({ theme }) => theme.accordion.labelFontWeight};
+  padding: ${({ theme, mods }) => mods.small ? theme.spacing.medium : theme.spacing.large};
+  color: ${({ theme, mods }) => mods.small ? theme.colors.black : theme.colors.primary};
+  font-family: ${({ theme }) => theme.fonts.brand};
+  font-size: ${({ mods }) => mods.small ? '14px' : '1.5em'};
```
tato123: It seems like it would be helpful to pull the font sizes from a set of font sizes in the theme. This would ensure that fonts are consistent with our grid system.

src/components/Accordion/Accordion.js
```diff
@@ -32,7 +40,7 @@ const ModdedIcon = styled(Icon)`
   &:after {
     padding-top: 0;
     padding-bottom: 0;
-    font-size: ${({ theme }) => theme.accordion.iconFontSize};
+    font-size: ${({ mods }) => mods.small ? '14px' : '1em'};
```
tato123: Same comment as Label, pulling `14px` and `1em` from a theme value would be preferrable

src/components/Button/Button.js
```diff
@@ -1,22 +1,24 @@
...
+  font-weight: 700;
+  font-family: ${({ theme }) => theme.fonts.brand};
+  text-transform: uppercase;
 
   background: ${({ theme }) => theme.colors.primary};
   border: 1px solid ${({ theme }) => theme.colors.primary};
   color: ${({ theme }) => theme.colors.white};
   border-radius: 3em;
-  padding: ${({ theme }) => theme.button.padding};
+  /* subtracting 1 from the design system height due to the border */
```
tato123: Comment says we are subtracting 1, but the code statically sets the values, unclear what the logic should be?

src/components/Button/Button.js
```diff
@@ -33,31 +35,31 @@ const ButtonImpl = styled.button`
...
 
   &:hover:not(:disabled), &:focus {
-    background-color: ${({ theme }) => theme.button.activeBG};
-    border-color: ${({ theme }) => theme.button.activeBG};
-    color: ${({ theme }) => theme.button.activeFG};
-    box-shadow: 0 2px 4px ${({ theme }) => theme.shadow.color};
+    background-color: ${({ theme }) => theme.colors.primaryDark};
+    border-color: ${({ theme }) => theme.colors.primaryDark};
+    color: ${({ theme }) => theme.colors.white};
+    box-shadow: 0 2px 4px rgba(0, 0, 0, 0.14);
```
tato123: A nearly identical drop shadow of `rgba(0,0,0,0.14)` appears in the baseTheme under `shadows.default`. Should we extract the color into its own value?

src/components/Code/Code.js
```diff
@@ -1,22 +1,24 @@
...
 const CodeImpl = styled.pre.withConfig({ displayName: 'Code' })`
-  font-family: ${({ theme }) => theme.inlineCode.fontFamily};
-  fontSize: ${({ theme }) => theme.inlineCode.fontSize};
-  background: ${({ theme }) => theme.inlineCode.bg};
-  color: ${({ theme }) => theme.inlineCode.fg};
-  border: ${({ theme }) => theme.inlineCode.border};
-  padding: ${({ theme }) => theme.inlineCode.padding}
-  borderRadius: ${({ theme }) => theme.inlineCode.borderRadius};
+  font-family: ${({ theme }) => theme.fonts.monospace};
+  fontSize: 0.85em;
```
tato123: FYI: `em` units are relative to their parent, just want to make sure that is the intention for this components font-size. 

Also, this should be `font-size` and not `fontSize`

src/components/Code/Code.js
```diff
@@ -1,22 +1,24 @@
...
+  background: #f1f1f1;
+  color: ${({ theme }) => theme.colors.black};
+  border: 1px solid #e4e4e4;
+  padding: 0.3em;
+  borderRadius: 3px;
```
tato123: I believe this should be `border-radius` not borderRadius

src/components/Table/SimpleTable.js
```diff
@@ -47,11 +48,14 @@ class SimpleTable extends React.Component {
   };
 
   static defaultProps = {
-    renderRow: (item) => (
+    renderRow: item =>
       <Table.Row key={JSON.stringify(item)}>
-        {Object.values(item).map((val) => <Table.Cell>{defaultValueRenderer(val)}</Table.Cell>)}
-      </Table.Row>
-    ),
+        {Object.values(item).map(val =>
```
tato123: FYI: Lodash checks used above but es6 was used for this check, could use `_.values` for consistency (not really a big issue, just a heads up)

a-type:
Yeah, the spacing has not been addressed well from the beginning. My general direction is to begin ripping out any spacing on individual elements and expecting users to utilize the `Spacing` component with its named sizes.

I'm starting to feel more draw toward a second library which composes multiple components into packaged 'widget' like solutions (still not implementing any logic, but providing a more complex presentational out-of-the-box experience). These could include spacing since they're more targeted and opinionated. But for things like headers, etc... best to just give them 0 margin at all and let the user wrap them the way they're needed.

a-type:
Thanks for reviewing the CSS; most of that was just copied from the theme directly, I didn't really look much at the values as I went along.

src/components/Button/Button.js
a-type: The design system is a reference to the 'spec' at https://dev.bandwidth.com/design-system, not any value present in the code. I just wanted to leave a note about why it's 12px instead of 14px in case anyone saw it and wanted to 'fix' it. Though the comment itself is not that clear. It's copied over from the old theme files.

PR #26, opened 2017-09-21
KAT-1895 add show/hide by default in password field [2.0.7 - patch]

tato123:
- Adds a show / hide option by default for inputs of type `password`
- Developers can disable show/hide by using flag `disableShowPassword`
- Updated docs

### DIFF
--- a/src/components/Input/Input.js
+++ b/src/components/Input/Input.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled, { css } from 'styled-components';
+import Anchor from '../Anchor';
 
 const StyledInput = styled.input`
   color: ${({ theme }) => theme.colors.black};
@@ -15,6 +16,7 @@ const StyledInput = styled.input`
   padding: ${({ theme }) => theme.padding.medium};
   display: block;
   border: ${({ theme }) => theme.input.border};
+
 
   &:focus {
     box-shadow: inset 0 -5px 0 ${({ theme }) => theme.colors.primaryLight};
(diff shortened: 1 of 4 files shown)


src/components/Input/Input.js
```diff
@@ -138,7 +153,33 @@ class Input extends React.Component {
...
+  renderPasswordField = () => {
+    const toggleState = () => this.state._type === 'password' ? 'text' : 'password';
+    const handleClick = (evt) => {
+      evt.preventDefault();
+      this.setState({
+        _type: toggleState(this.state._type)
+      });
+    }
+
+    const getText = () => {
```
a-type: This function name seems kind of ambiguous for an input.

src/components/Input/Input.js
```diff
@@ -138,7 +153,33 @@ class Input extends React.Component {
...
+      return this.state._type === 'password' ? 'Show' : 'Hide'
+    }
+
+    return (
+      <div style={{position:'relative'}}>
+          <div style={{position:'absolute', right: '10px', top: '25%'}}>
```
a-type: Didn't feel like using styled-components for these? I can certainly see why, just curious.

src/components/Input/Input.js
tato123: Yup, you're right, that was just laziness on my part. Fixed

src/components/Input/Input.js
tato123: This was honestly just carelessness on my part leaving this in while I played around with it. Changed to styled-components

PR #31, opened 2017-10-02
Tabs component

ssuchkov:


src/components/Tabs/Tab.js
```diff
@@ -0,0 +1,117 @@
...
+
+    if (this.props.onClick) {
+      this.props.onClick(e);
+    }
+  }
+
+  render() {
+    const {isActive, displayName, disabled, vertical} = this.props;
+
+    return (this.props.displayContentOnly) ?
```
a-type: This is pretty confusing; based on one passed property, this component entirely changes its abstraction and interaction?

src/components/Tabs/Tab.js
```diff
@@ -0,0 +1,114 @@
...
+    if (this.props.onClick) {
+      this.props.onClick(e);
+    }
+  };
+
+  render() {
+    const { active, disabled, vertical } = this.props;
+
+    return (
+      <Div vertical={vertical}>
```
a-type: `Div` isn't a very descriptive component name. Also, it doesn't seem necessary; can't its styles be moved to `TabContainer`?

src/components/Tabs/Tab.js
```diff
@@ -0,0 +1,114 @@
...
+Tab.defaultValues = {
+  disabled: false,
+  active: false,
+};
+
+Tab.propTypes = {
+  active: PropTypes.bool,
+  vertical: PropTypes.bool,
+  handleSelect: PropTypes.func,
+  namespace: PropTypes.string,
```
a-type: What is `namespace`? Documentation on the propTypes might be helpful.

src/components/Tabs/TabContainer.js
```diff
@@ -0,0 +1,45 @@
...
+  ${({ vertical }) => {
+    if (vertical) {
+      return css`  
+    width: 100%;
+    display: table;
+  `;
+    }
+  }}`;
+
+export default class TabContainer extends React.Component {
```
a-type: `TabContainer` component name collides with `TabContainer` internal component under `Tab`. Need to be careful with generic names like `Container`. This appears to be an element that contains tabs; perhaps `TabList` is a better name?

src/components/Tabs/TabContainer.js
```diff
@@ -0,0 +1,45 @@
...
+     */
+    children: PropTypes.node,
+    /**
+     * Enable vertical tab layout
+     */
+    vertical: PropTypes.bool,
+  };
+
+  attachParams = (children, vertical) => {
+    if (typeof children !== 'object') {
```
a-type: This check is kind of confusing, and I don't think it should be necessary. `React.Children.map` automatically casts children to an array.

src/components/Tabs/Tab.js
```diff
@@ -0,0 +1,114 @@
...
+    width: calc(100% + 1px);
+  `;
+      }
+    }
+} `;
+
+export default class Tab extends React.Component {
+
+  handleClick = (e) => {
+    this.props.handleSelect(this.props.namespace, this.props.index);
```
a-type: I don't think the `index` needs to be passed down this far. `onClick` would be a better prop to use. Pass a pre-bound handler function to `onClick` that is already bound to the tab's index and whatever other information you need about the tab.

```
createTabClickHandler = (index) => () => onTabSelected(index);

<Tab onClick={createTabClickHandler(tabIndex)} />
```

That's pretty easy to roll into the tab rendering logic and keeps the Tab itself free of extraneous data. I'd like to see Tab reduced to a simple `styled-components` component, rendering logic only.

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,115 @@
...
+      <TabsDiv vertical={this.props.vertical}>{tabs}</TabsDiv>
+    );
+  }
+}
+
+Tabs.defaultProps = {
+  vertical: false,
+};
+
+Tabs.usage = `
```
a-type: `.usage` is deprecated; with our new documentation you include a `.md` file next to your component (like `Tabs.md`)

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,115 @@
...
+        name: this.props.name,
+        vertical: this.props.vertical,
+      });
+    return (
+      <TabsDiv vertical={this.props.vertical}>{tabs}</TabsDiv>
+    );
+  }
+}
+
+Tabs.defaultProps = {
```
a-type: `defaultProps` is better defined as a static value next to `propTypes` (`static` is the same as assigning the value as a property after the class is defined).

src/components/Tabs/Tab.js
ssuchkov: Yes, sounds good
Also I removed namespace since tab group's names could be managed in higher components

src/components/Tabs/Tab.js
ssuchkov: fixed

src/components/Tabs/Tab.js
ssuchkov: firstly that was the way to detect which tab section is clicked 
removed

src/components/Tabs/TabContainer.js
ssuchkov: for sure that could help to get rid of collisions

src/components/Tabs/TabContainer.js
ssuchkov: done

src/components/Tabs/Tabs.js
ssuchkov: done

src/components/Tabs/Tabs.js
ssuchkov: got it

src/components/Tabs/Tab.js
```diff
@@ -72,43 +67,52 @@ const TabContainer = styled.label`
...
 
-    if (this.props.onClick) {
-      this.props.onClick(e);
-    }
+  static defaultProps = {
+    disabled: false,
+    active: false,
   };
 
+  createTabClickHandler = (index) => () => this.props.onTabSelected(index);
```
a-type: Getting there, but one last step could make this component just a simple styled-components export... You can move this handler creator out into the `renderTabs` loop in `Tabs` and pass it directly to the `onClick` handler on `Tab`. If you do that, you don't need to define a component at all in this file, just export the one created by `styled.div`.

The reason I'm pushing this is that one of the big objectives of this library is to create tools which are as versatile as possible. The more I've expanded and worked on it, the more I find that separating out the visual components from the behavioral ones achieves that goal. In this case, a simple `div` styled like a tab which has no special requirements will be more flexible and useful than one which is coupled to an implementation (that is, one that requires an index to be provided). By moving all of the implementation details into a behavioral component (in this case, `Tabs`), the visual component can be used in a wider variety of circumstances (for instance, if you just wanted to render a visual tab in another component later but didn't need tab-switching functionality).

src/components/Tabs/Tabs.md
```diff
@@ -0,0 +1,48 @@
+```javascript
```
a-type: Nice example documentation!

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,86 @@
...
+  };
+
+  renderTabs = (children,
+                { onTabSelected, selectedTabIndex, vertical }) => React.Children.map(children, (child, index) => {
+      if (!React.isValidElement(child)) {
+        return child;
+      }
+
+      // rendering Tab
+      return (
```
a-type: Based on the usage, it looks like you pass in Tab components as children, but then you wrap them in Tab again here. Won't that result in double-nested Tab structure?

I can see that you want this component to provide the `onClick` handler, `active`, `disabled`, `vertical` for child tabs. You could use `React.cloneElement` to provide those props to the Tab children that are already there.

ssuchkov: We do not have double-nested structure since we used child.props.children means we ignore the tag itself and replace it by Tab, which is not the best approach I believe

Switching to React.cloneElement

a-type:
@ssuchkov Released in v2.0.20

PR #40, opened 2017-10-27
z-ordering fix for sidebar list items

a-type:
Sidebar list items don't set a positive z-index value, but they are designed to overlap the dividing line of the SidebarLayout component. If a user makes the content of a SidebarLayout `position: relative`, or otherwise elevates its z-ordering, the SidebarListItem will no longer overlap it properly. Setting a positive z-index value maintains the behavior. 100 was chosen arbitrarily... It might be nice to establish a guideline system for z-index values if we deal enough with them in the future.


mowoc-bw:
Thus begins the z-index wars

PR #45, opened 2017-11-06
KAT-1960 add height as a prop to fields and textarea

lunxiao:


src/layouts/Flow/fields/flowField.js
```diff
@@ -89,6 +89,7 @@ export default (FieldComponent) => class FlowField extends React.Component {
       choices,
       description,
       type,
+      height,
```
a-type: `flowField` is generic and doesn't need to be concerned with field-specific props like height, so it shouldn't be added here. When you pass height to Flow.field.TextArea, you can use extraProps.

```
<Flow.field.TextArea extraProps={{ height: '100px' }} />
```
or
```
<Field component={Flow.field.TextArea} extraProps={{ height: '100px' }} />
```

PR #46, opened 2017-11-07
Integrate a library for themeing solutions

a-type:
So I made this thing, [react-studs](https://github.com/a-type/studs). I've gone and integrated it into the SCL to see how it fares in a real-world scenario. So far I'm fairly pleased with it, but I do worry a bit about the perf.

For more information about studs, see the link above, which should link to docs.

At a high level, this PR introduces the ability to specify customizable attributes for all components that we can then override at-will in our apps. It also simplifies the purpose of themes to encompass only coloring and other tweakable values. It splits our themes into `iris` and `catapult`, based on colors used.

I've already tested this by migrating Catapult on a local branch. It seems seaworthy.

### Preview

See an uploaded copy of the docs here: https://s3.amazonaws.com/gforrest/shared-components/studs/index.html


a-type:
Concerns presented are wide enough that I think we should take a different approach. I'll close this PR and submit a new one soon.

PR #47, opened 2017-11-15
Customized

a-type:
This is a second pass at making the library more customizable, themeable, and easier to use.

There's no special libraries at play here. The basic strategy is:

1. Separate presentational components from behavioral components. Presentational components go in `/styles`.
2. Make all sub-components into props. So if you want to override a presentational component, you just create your own version of it and pass it in as a prop.

This finally gets us using styled-components the way it was meant to be used: for visual primitives. The styled-components library authors repeatedly stress that users should try to restrict usage of styled-components to the 'leaf' nodes in the render tree. We're much closer to that ideal now.

In addition to reworking the structure, I've also updated styling to be more consistent with the design system on many components.

In addition, I've added a new Iris theme which updates the colors.

You can see a hosted version here: https://s3.amazonaws.com/gforrest/shared-components/customized/index.html


a-type:
No, you don't necessarily have to read all the code here. There's a ton of changes.

Definitely pick through the hosted version. And read one or two component diffs to see the big idea.

tato123:
I found a couple of minor bugs, I'm unsure if these are just general SCL bugs or introduced as part of the fix:

- Navigation items dissappear when swithcing between themes (light navigation)
- RadioGroup with descriptions, changing theme causes layout to randomly change
- Select component x marks are off-centered vertically

a-type:
@tato123 Good catch on those. I've got a fix for the navigation item color and the select centering. I think the radio group layout changes because the lorem ipsum content text gets regenerated every time it renders.

a-type:
Updated the hosted version. Reveiwers, please be sure to check out the new Navigation usage, since it's changed dramatically.

https://s3.amazonaws.com/gforrest/shared-components/customized/index.html#navigation

a-type:
@tato123 good call on SearchBox, I'll fix that. The Modal doesn't close when you click the shadow, but does expose props you could use to do that.

PR #50, opened 2017-11-30
Note component

a-type:
Why `Note` and not `Comment`? While looking through the CSS in the design system, I noticed the sticky-note-like element is actually called "note". I think it's more useful and idiomatic as a name.

## PR Checklist

## Breaking Changes

* None

* None

## New Visual Components

* NoteBody: renders the outer body of a note component
* NoteCorner: renders the corner graphic on a note component

## New Behavioral Components

* Note: places children inside a graphical note container


src/components/Note/Note.js
```diff
@@ -0,0 +1,33 @@
...
+
+Note.propTypes = {
+  /**
+   * Contents of the note
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * Override the component that renders the body container
+   */
+  Body: PropTypes.func,
```
tato123: `Body` seems like it could be confused with HTML `body`. I don't have a better suggested name but it does seem like it could be confused

a-type: Good point. I could go back to the typical 'Container' naming, but that seems unintuitive here.

PR #51, opened 2017-11-30
Adding Breadcrumbs component

tbowenbw:
## PR Checklist

## Breaking Changes

N/A

N/A

## New Visual Components

* Breadcrumbs - container for individual Breadcrumb components.  Each Breadcrumb is separated by a ">", per the design system.
* Breadcrumb - individual nav items commonly seen at the top of webpages to represent where the user is located in the webapp.  

## New Behavioral Components

N/A

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.js
+++ b/src/components/Breadcrumb/Breadcrumb.js
@@ -1,6 +1,6 @@
 import styled from 'styled-components';
 
-const Breadcrumb = styled.div`
+export default styled.div`
   float: left;
   
   &>* {
@@ -11,5 +11,3 @@ const Breadcrumb = styled.div`
    clear: both;
   }
 `;
-
-export default Breadcrumb;
--- a/src/components/Breadcrumb/Breadcrumb.js
+++ b/src/components/Breadcrumb/Breadcrumb.js
@@ -6,6 +6,10 @@ const Breadcrumb = styled.div`
   &>* {
     float: left;
   }
+  
+  &:after {
+   clear: both;
+  }
 `;
 
 export default Breadcrumb;
--- a/src/components/Breadcrumb/Breadcrumb.js
+++ b/src/components/Breadcrumb/Breadcrumb.js
@@ -1,0 +1,11 @@
+import styled from 'styled-components';
+
+const Breadcrumb = styled.div`
+  float: left;
+  
+  &>* {
+    float: left;
+  }
+`;
+
+export default Breadcrumb;
--- a/src/components/Breadcrumb/Breadcrumbs.js
+++ b/src/components/Breadcrumb/Breadcrumbs.js
@@ -1,0 +1,9 @@
+import styled from 'styled-components';
+
+const Breadcrumbs = styled.div`
+  &>*:not(:last-child)::after {
+    content: "\00a0 \00a0 ›\00a0 \00a0"
+  }
+`;
+
+export default Breadcrumbs;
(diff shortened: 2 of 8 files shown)


src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,11 @@
+import styled from 'styled-components';
+
+const Breadcrumb = styled.div`
+  float: left;
```
tato123: `float` modifies the normal page flow, you should also include a clearfix to reset the flow afterwards

```
&:after {
   clear: both;
}
```

src/components/Breadcrumb/Breadcrumbs.md
```diff
@@ -0,0 +1,15 @@
+```Breadcrumbs``` is intended to hold one or more ```Breadcrumb``` components.  This provides an easy way to structure
```
tato123: For in-line code you can just use a single tick, e.g. \`Breadcrumbs\`

tbowenbw:
@a-type I'm not sure how much the redundancy is an issue.  At the very least, that pattern should be similar to `<ul>` or `<li>`, which only allows for `<li>` elements as its direct children.  

I'm curious what internalizing some of that would look like.  Would we just assume each child element is a separate item in the breadcrumbs?

a-type:
@tbowenbw `ul` has the advantage of being a part of HTML standards, which produces warnings or errors if not used correctly. But it also has the disadvantage of not being a component, so it couldn't convert its children into `li` even if it wanted to.

The gist would be like so:

```js
/* inside Breadcrumbs React class */

render() {
  const { children } = this.props;
  // BreadcrumbsList is a styled-component, formerly Breadcrumbs
  return (
    <BreadcrumbsList>
      {React.children.map(children, child => <Breadcrumb>{child}</Breadcrumb>)}
    </BreadcrumbsList>
  );
}
```

`Breadcrumbs` would be an injectable prop here, so the user could override it if they wanted to wrap children in a different component.

I basically suggest this approach because I don't see any valid usecase of Breadcrumbs where you wouldn't want all of its children to be Breadcrumb components... so perhaps we can save the user the trouble of providing them.

tbowenbw:
@a-type I think we can have the best of both worlds here.  If I add a proptypes check on children, I could validate that each child is in fact a `Breadcrumb`.  Then, the developer gets to see an explicit error if they try to use anything other than a `Breadcrumb`.

The developer still has to explicitly define `Breadcrumb` elements, but I personally find that to be preferable to mapping each child to a `Breadcrumb`.  Given the use case of a developer who wants a link and an image in one nav item, `Breadcrumb` is more semantic than using a `div`.  I admit that the majority use case is just going to be `Anchor`s or plain text, which would require more verbose code.  The tradeoff seems to be convenience versus convention.

a-type:
@tbowenbw good reasoning about including multiple components. Your approach sounds good to me.

src/components/Breadcrumb/Breadcrumbs.js
```diff
@@ -0,0 +1,25 @@
+import styled from 'styled-components';
+import Breadcrumb from './Breadcrumb';
+import dedent from 'dedent';
+
+const Breadcrumbs = styled.div`
+  &>*:not(:last-child)::after {
```
a-type: Since we're already being specific about child component type, we can use the more efficient class name selector `${Breadcrumb}` here instead of `*`.

PR #52, opened 2017-11-30
Steps component

a-type:
## PR Checklist

## Breaking Changes

* None

* None

## New Visual Components

* StepContainer: renders the outer container of a step
* StepNumber: renders a number for a step, supports rendering a check when `complete=true`
* StepContent: renders the internal content of a step
* StepList: renders a vertical list

## New Behavioral Components

* Step: A single step. Supports `number`, `complete`.
* Steps: A set of numbered steps. Numbers are automatically assigned.

### DIFF
--- a/src/components/Steps/styles/StepContent.js
+++ b/src/components/Steps/styles/StepContent.js
@@ -1,0 +1,10 @@
+import styled from 'styled-components';
+import StepContainer from './StepContainer';
+
+export default styled.div`
+  ${StepContainer} > & {
+    flex: 1 0 0;
+    margin-top: 0.5em;
+    margin-left: 1em;
+  }
+`;
--- a/src/components/Steps/Step.js
+++ b/src/components/Steps/Step.js
@@ -1,0 +1,60 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import StepContainer from './styles/StepContainer';
+import StepNumber from './styles/StepNumber';
+import StepContent from './styles/StepContent';
+
+export default class Step extends React.Component {
+  static propTypes = {
+    /**
+     * Content of the step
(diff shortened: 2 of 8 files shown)


src/components/Steps/styles/StepContent.js
```diff
@@ -0,0 +1,10 @@
+import styled from 'styled-components';
+import StepContainer from './StepContainer';
+
+export default styled.div`
+  ${StepContainer} > & {
```
tbowenbw: Would this still work if the user decided to plug in a custom StepContainer as the Container prop in the Step component?  It might be that decoupling these styles from their parent makes them more extensible.

a-type: As a rule of thumb, parents should determine the layout of their children (that being `margin`, `flex`, and any other properties which orient a child within its parent).

This syntax is equivalent to writing `& > ${Step}` within `StepContainer`, so it satisfies that rule: StepContainer is determining how its children should be aligned if they are `Step`s.

There are two advantages to writing this rule here instead:

1. It's easier for a reader to see the styling rules which apply to `Step`, without having to go to different files to determine relationships
2. To the best of my knowledge, this rule should persist even if an end user extends `Step`. Customized `Step` components would therefore be laid out in the same manner when contained in a vanilla `StepContainer`.

Finally, to address your actual question, yes, if the user extends `StepContainer`, they will invalidate this rule. However, since `StepContainer`'s purpose is to lay out `Step` components, it was probably their intention to change this behavior anyway.

I'm still back and forth on just defining a static class name for `StepContainer`, though. Then, no matter which component the user extended, the rule would continue to apply. I've decided to wait and see how usage develops.

PR #55, opened 2017-12-13
First pass at creating Cards/Card components

tbowenbw:
## PR Checklist

## Breaking Changes

No breaking changes.

No existing components are changed.

## New Visual Components

* Cards, a collection of Card components.
* Card, a styled container for other content.  Card components have an optional header which may contain a title for the card and may use a background image.
* CardSection delineates one section inside of a Card.  Content in the Card may be divided into sections as appropriate.  Sections within a Card are visually distinct.

## New Behavioral Components

No new behavioral components.


tbowenbw:
Is the usage pattern for Cards, Cards.Card and Cards.Card.Section appropriate?  Since Cards are intended to be grouped, and also intended to contain arbitrary sections, this feels like the best way to model it to me, but feedback is welcome.

a-type:
Looks like Breadcrumb also got pulled into this branch.

src/components/Cards/Card.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import themeGet from 'extensions/themeGet';
+import DefaultCardHeader from './CardHeader';
+import CardSection from './CardSection';
+
+const OuterBox = styled.div`
```
a-type: All presentational styling components should be defined in `./styles/` and injected as props.

src/components/Cards/CardHeader.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+import H3 from 'components/H/H3';
+import themeGet from 'extensions/themeGet';
+
+const OuterContainer = styled.div`
```
a-type: Ditto comments on Card.js, style components should be their own modules.

src/components/Cards/Cards.js
```diff
@@ -0,0 +1,39 @@
+import styled from 'styled-components';
+import themeGet from 'extensions/themeGet';
+import Card from './Card';
+
+const Cards = styled.div`
+  display: flex;
+
+  & > * {
+    margin: ${themeGet('spacing.medium')};
+    border-radius: 5px;
```
a-type: I think `Card` by itself should apply border radius and shadow even if it's not enclosed in `Cards`.

src/components/Cards/Cards.js
```diff
@@ -0,0 +1,39 @@
...
+    box-shadow: ${themeGet('shadows.long')};
+  }
+`;
+
+Cards.Connected = Cards.extend`
+  & > * {
```
a-type: I think these relationships may be better inverted and placed in `Card`'s styling components...

```
${Cards} > & {
  margin: 0;
  / * etc */
}
```

This keeps relevant CSS properties contained in the stylesheet of the component they apply to, and makes the Card component more easily customizable on its own.

tbowenbw: Re: usage
It would be great if we had a real-world example of where this was going to be used, this could help inform the design decisions.  I like the idea of making Card the primary component, since that makes the usage of a single Card on the page much easier.

I'm not sure about using a child component of Card as a parent of Card on the page, this seems a bit counter-intuitive to me.  Would it perhaps be better to include a separate element e.g. Deck to optionally contain Cards?  The downside to this is that outside of semantics, it doesn't show a relationship between the two components.  In general, I struggle with figuring out how to relate two components, since modeling relationships immediately puts me in an object-oriented state of mind.

Either way using Card as the top-level element seems to be a better solution than the Cards.Card solution, so I will start refactoring towards that goal.

tbowenbw: Another thought concerning the Header design: having a Card.Header as its own component lets the markdown be more semantic, but doesn't enforce a single header at the top of a card.  That is to say, the developer can include more than one header, and can include them anywhere in the stack of card sections.  I'm not sure that this is much of a concern, since ultimately it's on the developer to use the components appropriately in the first place, but including header title/image props does prevent the developer from using the header in a manner which we didn't intend it to be used.

a-type: Including a reference to another component as a static property, i.e. `Cards.Group`, doesn't necessitate that component being considered a 'child'; just a convenience reference. For another example, see `Accordion.Group`. We only need to consider which component will be referenced more often. Since a `group : item` relationship is `1 : x, x >= 1`, we can reasonably assume that Card will be referenced most frequently and should be the primary export to save keystrokes for the developer.

tbowenbw:
Refactoring such that Card is the primary component wound up greatly simplifying a lot of the CSS and let me remove several presentational components.  This is definitely the better usage pattern.

src/components/Cards/CardSection.js
```diff
@@ -0,0 +1,20 @@
...
+const CardSection = styled.div`
+  padding: ${themeGet('spacing.medium')};
+  ${CardSection}:not(:last-of-type) {
+    border-bottom-color: ${themeGet('colors.border.medium')};
+    border-bottom-width: ${themeGet('thicknesses.normal')};
+    border-bottom-style: solid;
+  }
+
+  ${CardSection} > * {
+    margin: 0;
```
a-type: I'm a little wary of this rule for children, could you describe your reasoning behind it?

src/components/Cards/styles/HeaderWrapper.js
```diff
@@ -0,0 +1,16 @@
+import styled from 'styled-components';
+import themeGet from 'extensions/themeGet';
+
+const HeaderWrapper = styled.div`
+  color: ${themeGet('colors.background.default')};
+  font-size: 1.2em;
+  font-weight: 500;
+  text-transform: uppercase;
+  background-image: url(${props => props.image || 'none'});
```
a-type: If I understand correctly, this component will contain a white-colored header. To avoid that header becoming invisible if an image is not provided, a background color could also be specified. Maybe `colors.primary.dark`.

src/components/Cards/CardSection.js
tbowenbw: This was caught in visual inspection; when a paragraph tag is used in the CardSection, the margin on the top and bottom throws off the spacing, as the left and right padding are smaller than the combined top and bottom padding and margin.

Other options are to remove the p tag from usage, or simply let the developer handle the margin.  Ideally, though, the padding is the only spacing around the content, regardless of whatever spacing is specified within the content itself.

src/components/Cards/styles/HeaderWrapper.js
tbowenbw: This is a really good idea

src/components/Cards/CardSection.js
a-type: Seems like we should have a `P` component that removes that browser-default margin if we want to use `<p>`. But I'm ok with this for now. We'll see where things go.

PR #56, opened 2017-12-13
BWDB-2717 fix button styles

lunxiao:
## PR Checklist

- [x ] Exported components are documented with examples

## Breaking Changes

## New Visual Components

## New Behavioral Components

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -45,7 +45,7 @@ const Button = styled.button`
 
   &:hover:not(:disabled),
   &:focus:not(:disabled) {
-    background-color: ${get('colors.primary.alternate')};
+    background-color: ${get('colors.primary.dark')};
     border-color: ${get('colors.primary.alternate')};
     box-shadow: ${get('shadows.short')};
   }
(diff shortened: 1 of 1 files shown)


src/components/Button/Button.js
```diff
@@ -54,12 +55,12 @@ const Button = styled.button`
     color: ${get('colors.text.disabled')};
     border-color: ${get('colors.background.disabled')};
     cursor: default;
+    opacity: 0.5;
```
tato123: Based on the original story it appears `hover` should be darker, however lowering the opacity would lighten on hover, here's the original spec: 

- Hover state does not use the 'primary dark' color, instead uses a slightly darker normal primary color.
- Disabled hover icons are not visible; they should be white.

lunxiao: lower the opacity is the requirement from the story. I have already showed the changes to Shaun and he approved the look.

src/components/Button/Button.js
```diff
@@ -42,6 +42,7 @@ const Button = styled.button`
 
   outline: none;
 
+
   &:hover:not(:disabled),
   &:focus:not(:disabled) {
     background-color: ${get('colors.primary.alternate')};
```
a-type: Based on discussion with the UX slack channel, this should be changed to `'colors.primary.dark'`.

PR #59, opened 2018-01-02
KAT-1974 remove red shadow of required input field for FF

lunxiao:
## PR Checklist

- [ x] Exported components are documented with examples

## Breaking Changes

* NONE

* Changed shadow for required Input component to fix red shadow in FF

### DIFF
--- a/src/components/Input/styles/InputStyles.js
+++ b/src/components/Input/styles/InputStyles.js
@@ -24,6 +24,10 @@ const InputStyles = styled.input`
     border-color: ${get('colors.border.medium')};
   }
 
+  &:required {
+    box-shadow: none;
+  }
+
   &:disabled {
     background: ${get('colors.background.disabled')};
     border-color: ${get('colors.border.medium')};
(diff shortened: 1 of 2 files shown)


src/components/Input/styles/InputStyles.js
```diff
@@ -24,6 +24,10 @@ const InputStyles = styled.input`
     border-color: ${get('colors.border.medium')};
   }
 
+  &:required {
+    box-shadow: none;
```
a-type: This removes the focus effect on Firefox as well by overriding the rule from line 22.

a-type: Probably needs to be moved above that rule.

a-type:
I'm concerned that this fix is not going to be very simple, considering it works on Chrome as intended already.

The purpose is not to remove the red effect entirely, only to keep it from appearing until the user has 'visited' the input and left it blank.

a-type:
Just noticed the default invalid state is actually broken in master right now. I have a fix ready: https://github.com/Bandwidth/shared-components/pull/61

src/components/Input/styles/InputStyles.js
lunxiao: Yeah, I changed the order. Other than that the fix does not change any existing behavior except for removing the red shadow on FF when required fields are initially displayed.

PR #60, opened 2018-01-02
BWDB-2769 fix h3 styles in SCL

lunxiao:
## PR Checklist

- [x ] Exported components are documented with examples
- x[ ] Props have JSDoc comments

Acceptance Criteria
The Heading 3 style should be defined by:

color: #666;
font-family: 'Overpass', sans-serif;
font-size: 1.75em; /* 14px base * 1.75 = 24.5px */
font-weight: 300;
letter-spacing: 0;
line-height: 1.25;

### DIFF
--- a/src/components/H/H3.js
+++ b/src/components/H/H3.js
@@ -6,9 +6,10 @@ import get from 'extensions/themeGet';
 const H3 = styled.h3.withConfig({ displayName: 'H3' })`
   color: ${get('colors.text.default')};
   font-weight: 300;
-  font-family: ${get('fonts.brand')};
+  font-family: 'Overpass', sans-serif;
   font-size: 1.75em;
-  line-height: 1em;
+  line-height: 1.25;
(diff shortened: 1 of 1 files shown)


src/components/H/H3.js
```diff
@@ -6,9 +6,10 @@ import get from 'extensions/themeGet';
 const H3 = styled.h3.withConfig({ displayName: 'H3' })`
   color: ${get('colors.text.default')};
   font-weight: 300;
-  font-family: ${get('fonts.brand')};
+  font-family: 'Overpass', sans-serif;
```
a-type: Let's keep this controlled by the brand font in the theme, in case we need to update the font across the library later.

lunxiao: brand fonts have more than what is needed here. Does it matter?

a-type: Any font after the first one is just a fallback, so it won't make any difference if things go well.

PR #62, opened 2018-01-03
KAT-2005 add optional text inside Input

lunxiao:
## PR Checklist

- x[ ] Exported components are documented with examples
- [ x] All relevant visual sub-components can be overridden via props

Add optional text that can be displayed inside text Input field and an associated function can be executed when the text is clicked.


a-type:
This implementation works perfectly well, but I think we may be better off doing something more flexible. What about a `controls` prop which accepts any arbitrary React node and renders it inline to the right of the input? We could then convert the password input to use `controls` under the hood.

Allowing any arbitrary content will help prepare us for more advanced use cases. I know the IP address grouping uses a variety of icon controls within its inputs.

```js
<Input
	value="foo"
	controls={(
		<div>
			<Icon name="delete" onClick={deleteField} />
			<Icon name="grab" onClick={moveField} />
		</div>
	)}
/>
```

PR #63, opened 2018-01-08
Fixing alert styles based on feedback

tbowenbw:
## PR Checklist

## Breaking Changes

None

* Alerts had their styling updated to match UX feedback
* Small versions of Alerts were also updated

## New Visual Components

None

## New Behavioral Components

None


a-type:
Looks like this PR doesn't include the docs build. Run `npm run build` and commit the resulting files to fix that.

src/components/Alert/styles/AlertBorder.js
```diff
@@ -3,15 +3,18 @@ import get from 'extensions/themeGet';
...
-  font-size: 1em;
-  line-height: 1em;
+  font-size: 14px;
+  line-height: 1.5em;
   position: relative;
   display: flex;
   flex-direction: row;
+  min-height: 53px;
+  letter-spacing: 0.02em;
+  white-space: wrap;
```
a-type: `wrap` isn't a valid `white-space` value. What were you intending with this?

src/components/Alert/styles/AlertBorder.js
```diff
@@ -36,8 +39,9 @@ const AlertBorder = styled.div`
 `;
 
 AlertBorder.Small = AlertBorder.extend`
-  padding: 0.5em 1em;
+  padding: 5px;
   font-size: 0.8em;
+  min-height: 30px;
```
a-type: Font weight should be `300` on the small alert according to the guidelines.

src/components/Alert/styles/AlertIcon.js
```diff
@@ -37,22 +38,39 @@ const AlertIcon = styled.i`
...
   ${AlertBorder} > & {
-    margin: 0 1em auto 0;
+    margin: 0 10px auto 0;
     flex: 0 0 auto;
   }
 `;
 
+const AlertIcon = getAlertIcon(SIZE);
+AlertIcon.Small = getAlertIcon(SMALL_SIZE).extend`
+  font-size: 1.25em;
```
a-type: Small icon font size should be `0.8em`.

src/components/Alert/styles/AlertBorder.js
```diff
@@ -3,15 +3,18 @@ import get from 'extensions/themeGet';
 
 const AlertBorder = styled.div`
   padding: ${get('spacing.small')} ${get('spacing.medium')};
-  font-weight: 500;
+  font-weight: 200;
   border-width: ${get('thicknesses.normal')};
   border-style: solid;
```
a-type: I noticed when working locally with this branch that the `border-color` is not defined, which means border will take its color from the `color` css attribute by default. That was probably an oversight. According to the guidelines, it should have an explicit color based on alert type, `#00bef0` (`colors.primary.default`) for info, `#00fbb9` (I actually think we should replace `colors.positive.border` with that value) for success, `#ffb39e` (again, replace `colors.negative.border` with this value) for error.

For context, the values in the theme were taken by me from our company's official brand guides verbatim. I think it's proper for us to replace them with new values when we get them from UX.

PR #64, opened 2018-01-16
BWDB-2888 update peerDependency so that react v16 is in the picture

lunxiao:

updated peerDependecy so that React v16 is allowed

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,8 +37,8 @@
   },
   "homepage": "https://github.com/Bandwidth/shared-components#readme",
   "peerDependencies": {
-    "react": "15.5.x",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -37,8 +37,8 @@
   },
   "homepage": "https://github.com/Bandwidth/shared-components#readme",
   "peerDependencies": {
-    "react": "15.5.x",
-    "react-dom": "15.5.x"
+    "react": ">=15.5.x <= 16.0.0",
```
a-type: Can we make this `16.0.x` to give ourselves some flexibility to use versions of 16 which have patch changes (usually bugfixes or security issues)?

PR #75, opened 2018-03-19
BWDB2955 added new styles + color & size variants

zelitzsi:
## PR Checklist

## Breaking Changes

* Potentially some minor color changes from changing the Iris theme; this is mostly in changing the disabled colors, which we want to be consistent across the application

* Altered `BulkSelect` and `BulkSelectItem` color scheme, in particular when disabled
* `BulkSelectItem` is now a button, which supports disabled behavior unlike `div`

## New Visual Components

* Added new color variant of `BulkSelect`

## Other
* Fixed Catapult theme provider not being set properly

### DIFF
--- a/src/components/BulkSelect/BulkSelect.js
+++ b/src/components/BulkSelect/BulkSelect.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { withProps } from 'recompose';
 import theme from '../../theme';
 import styled from 'styled-components';
 import BulkSelectItem from './styles/BulkSelectItem';
@@ -97,6 +98,7 @@ class BulkSelect extends React.Component {
             <Item
               key={computeItemKey(item.data)}
               selected={item.selected}
+              disabled={item.disabled}
               onClick={this.createItemClickHandler(item)}
             >
               {resolvedRenderItemContents(item.data, item.selected)}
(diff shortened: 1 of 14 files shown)


zelitzsi:
bulkselect

src/components/BulkSelect/BulkSelect.js
```diff
@@ -114,4 +116,10 @@ class BulkSelect extends React.Component {
   }
 }
 
+BulkSelect.Small = withProps({ ItemContainer: BulkSelectItemContainer.Small })(
+  BulkSelect,
+);
+
+BulkSelect.Variant = withProps({ Item: BulkSelectItem.Variant })(BulkSelect);
```
a-type: Variants should have descriptive names.

src/theme/irisTheme.js
```diff
@@ -55,17 +57,20 @@ export const colors = {
   graphics: ['#41d3bd', '#a95adf'],
   border: {
     medium: '#c2c2c2',
+    primary: '#004658',
```
a-type: As a consequence of the lack of organization in the theme, `colors.primary.border` already exists. Not that I disagree with moving the value here, but we do now have two different color values for `colors.primary.border` and `colors.border.primary`.

Trouble is, some of our colors are organized by color meaning ('primary'), some by usage ('border'). I guess this could all use some work.

zelitzsi: Changed to `colors.border.dark`.

PR #85, opened 2018-04-02
empty labels and help text rows in fields are handled correctly

zelitzsi:
## PR Checklist

* Formerly, if a `<Field>` had no label or help text, those grid areas would be empty but would still exist, so there would still be spacing in the gap between those empty sections and the content. Now, we remove those sections if they are not utilized by any children. This should make it easier to line up fields.
* Increased `<Input />` border size to 2px
* Modified `<Button />` padding
* Fixed `<ToggleInput />` sizing to not include extra sizing from being inline
* Grid items should now have 30px between rows


src/layouts/Fields/styles/FieldRow.js
```diff
@@ -8,33 +8,66 @@ const copyArea = (input, times) =>
...
+          'spacing.large',
+        )({ theme })}`,
+        // each column has an equal size. To make fields larger than
+        // adjacent fields, use the columnSpan prop on Field to span
+        // multiple columns. This keeps all field sizes directly proportional
+        gridTemplateColumns: `repeat(${columns}, 1fr)`,
+        // label and helpText rows expand or contract as needed.
+        // content row has a minimum size of 53px, and can expand as needed.
+        gridTemplateRows: `${
+          hasLabel(children) ? 'auto' : ''
```
a-type: This logic starts to get a little confusing, since the logic here must align with the logic for creating template areas. I'd almost rather abandon the use of `grid-gap` altogether in CSS and move to include the spacing within the label/helpText elements themselves to avoid all this conditional processing. What do you think?

zelitzsi: I don't think that would work too well for two reasons:
- When creating gaps in a series of items, you can use `:last-child` to make sure the final one doesn't add more space, but that isn't going to work in two dimensions very well.
- Instead, we could use gaps between fields but consider the full label + content + help text to be a single field. Part of the benefit we get out of css grid is making sure everything lines up very consistently. We could probably do it by having conditional spacing I guess.

I might be able to clean up the logic further - I can use one function to decide which areas should be present, then divert to different functions to map them to strings.

PR #88, opened 2018-04-05
Multi select

zelitzsi:
## PR Checklist

* **Accordions** can now be set to unmount their contents when the **Accordion** is closed. This significantly improves performance in cases where a large number of DOM elements are hidden.

## New Visual Components

* Added new **DragBox** component. It allows for components inside an area to be drag selectable. It has some behavioral elements, but I think it makes sense in the context of the component.
* Added **DragBox.Select** component. It effectively is just a convenient implementation of **DragBox** + **Selectable**, and it allows drag select functionality to be very simple and straightforward to add to pretty much any kind of element.
* Added **ToggleButton**. This is just **BulkSelectItem** moved out into its own component, as the new **Selectable** behavior largely eliminates the need for **BulkSelect** at all (outside of formatting).

## New Behavioral Components

* Added new **Selectable** behavior. This behavior simply keeps track of items as they are selected and deselected, and provides functions in its render prop to be able to control the selection. It would be simple to expand this component in the future to support other things like exclusivity.

## Notes
* I tried to support scrolling in **DragBox**; it works if page is being scrolled but not if an element inside the page is being scrolled. I'm not sure how to best address that problem. We can attach a listener to an onScroll value, but the element we would want to attach it to will be much higher in the DOM tree.


src/behaviors/Selectable/Selectable.js
```diff
@@ -0,0 +1,61 @@
...
+  state = { selected: new Set() };
+
+  static defaultProps = {
+    render: noop,
+  };
+
+  selectItem = key => {
+    const { state: { selected } } = this;
+    selected.add(key);
+    this.setState({ selected });
```
a-type: This should be a function-based state update to avoid race conditions. Playing around with this locally, I was able to click items fast enough that the state could not keep up.

src/components/DragBox/DragBox.js
```diff
@@ -0,0 +1,271 @@
...
+    if (!shouldDrawRect(rect)) return;
+    return renderRect({
+      left: rect.x,
+      top: rect.y,
+      width: rect.width,
+      height: rect.height,
+    });
+  };
+
+  getRef = key => el => (this.childElements[key] = el);
```
a-type: refs are called with `null` when the element unmounts. I suppose it's kind of a minor memory leak to leave references to null items in the child map... this could possibly be extended to cleanup references when it receives null. Up to you.

src/components/DragBox/DragBox.js
```diff
@@ -0,0 +1,271 @@
...
+      onCollisionEnd,
+      collisions,
+      getRef,
+    });
+  };
+
+  render() {
+    const { onMouseDown, renderRect, renderContents } = this;
+    return (
+      <div ref={el => (this.dragElement = el)} onMouseDown={onMouseDown}>
```
a-type: ref could be moved to class function, or you could use `React.createRef` to do refs the recommended 'clean' way (I just learned about this today) https://reactjs.org/docs/refs-and-the-dom.html

src/behaviors/Selectable/Selectable.js
zelitzsi: I'm not really sure what you mean by function-based state update.

src/components/DragBox/DragBox.js
zelitzsi: That's really good to know, I knew there was a memory leak here but I wasn't sure how to fix it.

src/behaviors/Selectable/Selectable.js
a-type: You can call `this.setState(currentState => ({ selected: currentState.selected.add(key) });` and it will enqueue that modification as a discrete one on the internal buffer, so it will avoid conflicting with other concurrent state changes.

I don't think the built-in `Set`'s `.add` returns the set, though... for that reason I usually prefer to import `immutable`'s `Set`.

zelitzsi: `Set.add()` does but `Set.delete()` does not.

src/components/DragBox/DragBox.js
```diff
@@ -311,10 +323,14 @@ class DragBox extends React.Component {
   render() {
     const { refDragElement, onMouseDown, renderRect, renderContents } = this;
     return (
-      <div ref={refDragElement} onMouseDown={onMouseDown}>
+      <DragBoxDiv
+        ref={refDragElement}
+        onMouseDown={onMouseDown}
+        disablePointerEvents={this.shouldDrawRect(this.calcRect())}
```
a-type: This seems a little wasteful since the same functions are being called within `renderRect`. Could `renderRect` be updated to accept a rectangle calculated in the scope of this function?

src/components/BulkSelect/BulkSelect.js
```diff
@@ -3,11 +3,15 @@ import PropTypes from 'prop-types';
...
+import ToggleButton from '../ToggleButton';
 import BulkSelectBorder from './styles/BulkSelectBorder';
 import BulkSelectDivider from './styles/BulkSelectDivider';
 import BulkSelectItemContainer from './styles/BulkSelectItemContainer';
 
+/**
+ * @deprecated BulkSelect has been deprecated in favor of using [Grid](/#!/Grid) with
+ * [ToggleButton](/#!/ToggleButton), optionally with [Selectable](/#!/Selectable) and/or [DragBox](/#!/DragBox).
+ */
 class BulkSelect extends React.Component {
```
tato123: Would it make sense to include a debug or warning log or something to indicate that this component has been deprecated? (since it's something that React does)

src/components/DragBox/DragBox.js
```diff
@@ -0,0 +1,351 @@
...
+    ),
+  };
+
+  childElements = {};
+
+  componentDidMount = () => {
+    const el =
+      window.document.querySelector(this.props.scrollSelector) ||
+      window.document;
+    el.addEventListener('scroll', debounce(this.handleScroll), 16);
```
tato123: Memory leak, `debounce` creates a new function wrapper. On `componentWillUnmount` we are removing `this.handleScroll` which is a different function reference. As a result the a DOM node will retain a reference to this object and the el domtree.

More succinctly:

Add a reference to `debounce(this.handleScroll)` use that reference for removing the event listener

src/components/BulkSelect/BulkSelect.js
tato123: I should have indicated this was an open question rather than a required change.

I was thinking potentially on initial creation of the component in a lifecycle method

zelitzsi: Okay - I'm a little concerned about spamming logs if I put it into the constructor, so I'm going to defer this for now.

PR #89, opened 2018-04-06
BWDB-3222 add new icons

lunxiao:
adding new icons


lunxiao:
Sounds good to me. Will do.

PR #94, opened 2018-04-10
Rework RadioButton style and usage

a-type:
gif

## PR Checklist

## Breaking Changes

* RadioGroup is now deprecated. RadioButton should be used directly, and over time we may be able to re-introduce some behaviors to make that more streamlined for forms, etc.
* RadioButton now accepts `children` instead of `content` for rendering arbitrary contents (this was not widely used, and RadioGroup has been updated to abstract it away).

* RadioButton has a new color scheme and styling rules
* RadioButton now exposes a `.Group` static property which can be used to visually group multiple RadioButtons together

## New Visual Components

* None

## New Behavioral Components

* None


src/components/RadioGroup/RadioButton.js
```diff
@@ -23,9 +28,15 @@ class RadioButton extends React.Component {
      */
     label: PropTypes.string.isRequired,
     /**
-     * Content to render inside the main part of the button.
+     * Content to render inside the main part of the button. You can pass
+     * a node, string, or a function. If a function is passed, it recieves
```
zelitzsi: sp: receives

src/components/RadioGroup/RadioButton.js
```diff
@@ -80,6 +91,25 @@ class RadioButton extends React.Component {
...
+  static Large = withProps({
+    Label: RadioGroupButtonLabel.Large,
+  })(RadioButton);
+
+  static Group = RadioGroupContainer;
+
+  renderChildren = () => {
+    const { children, checked } = this.props;
+
+    if (typeof children === 'function') {
```
zelitzsi: Does this actually work? I thought you needed to use the `React.children` API to handle children - I would expect this to be a weird wrapper around a function that you would need to unwrap with `React.children.only()`.

a-type: `children` is just whatever you pass it, right? I guess if you passed a component class into `children`, this would break. But as a render prop I think it works.

zelitzsi: It looks okay when I look at it in `console.log`. Maybe I'm remembering past React issues...

PR #98, opened 2018-04-13
icon docs now use a table

zelitzsi:
The icon docs were confusing before. Now they are arranged in a table.

icondocs


src/components/Icon/Icon.md
```diff
@@ -10,16 +10,23 @@ _Available Icons:_
...
+      </Table.Row>
     ))
-  }
-  </div>
+  }</Table.Small>
 );
-<IconList />
-```
 
+<IconList2 />
```
tbowenbw: Where does `IconList2` come from?

zelitzsi: Good catch, I was cleaning up some code and missed this.

PR #107, opened 2018-04-26
New Link

a-type:
## PR Checklist

## Breaking Changes

* Nothing breaking, but Anchor was deprecated

* Icon now defaults to `help` when the icon name supplied does not match anything.

## New Visual Components

* Three new Link components: Text, Icon, and Wrap. These are all distinct components (not props on one component).


lunxiao:
What are the reasons `Anchor` is deprecated? Anchor supports `to` and `onClick` which are used at the same time in number search. Does `Link` support `onClick`?

a-type:
@lunxiao The Anchor component is confusingly named (I've gotten feedback from other developers in the company), and the code is overly complicated. It has some bugs which would be difficult to fix without rewriting it. `Link` supports all props; it passes through any provided props to the underlying `<a>` element.

src/components/Link/linkExtensions.js
```diff
@@ -0,0 +1,19 @@
...
+ * function to the React-Router Link
+ */
+
+const isExternal = url => /^(https?:)*\/\//.test(url);
+
+export default WrappedLink => ({ to, children, ...rest }) => (
+  <WrappedLink
+    to={to}
+    {...(isExternal(to) ? { target: '_blank', rel: 'noopener' } : {})}
+    external={isExternal(to)}
```
zelitzsi: Does this attribute actually do anything? React Router seems to indicate that this will be passed through to the `<a>` tag, but I don't see docs for it in MDN either...

src/components/Link/linkExtensions.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+/**
+ * A HOC that provides a bit more intelligence / intuitive
+ * function to the React-Router Link
+ */
+
+const isExternal = url => /^(https?:)*\/\//.test(url);
```
zelitzsi: This is not actually an external url, this is an absolute one. It is possible to provide an internal URL with an absolute path. See https://stackoverflow.com/questions/904046/absolute-urls-relative-urls-and/904066#904066

src/components/Link/linkExtensions.js
a-type: It affects the rendering of the styled component. It won't be passed through to the underlying component below that.

src/components/Link/linkExtensions.js
a-type: While possible, seems unlikely based on our usage to date. But I could see an argument to remove heuristics and make the `external` prop explicit from the user. Opinions?

zelitzsi: I suppose this method is okay if we are using it purely for styling.

denysYefremov:
Will it have `innerRef`? Because not sure how it works for `Anchor` but I couldn't use it

zelitzsi:
Closed as it has sat for a long time.

PR #109, opened 2018-05-02
Add spacing prop support, align spacing on headings to baseline and cap

a-type:
headings

Pictured: the red lines are the boundaries of the element relative to other elements on the page. The blue dotted lines are 30px from the red lines.

## PR Checklist

## Breaking Changes

* Visual breaking change: Heading margin has been adjusted to collapse down to the baseline and cap line of the text. This may break some layouts.

* Added `spacing` prop to all H components, allowing customized spacing relative to the baseline and cap line.

### DIFF
--- a/src/extensions/userTextSpacing.js
+++ b/src/extensions/userTextSpacing.js
@@ -1,0 +1,90 @@
+/**
+ * EXPERIMENTAL
+ *
+ * Takes a user's `spacing` prop and converts it into margins
+ * which can be applied to text elements so that they are spaced
+ * from the cap-line and baseline, rather than the element bounds.
+ *
+ * This is accomplished by calculating margin with offsets of
+ * em values to 'dig' into the element a bit to reach the cap and baselines.
+ * These em values are based on visual inspection and would vary
+ * by font.
+ *
+ * Passing props as a whole so that we have access to dynamic theme
+ * information.
+ */
+
+import _ from 'lodash';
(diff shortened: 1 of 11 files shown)


src/extensions/userTextSpacing.js
```diff
@@ -0,0 +1,90 @@
...
+  /* these offset values collapse the outer element boundaries so that
+   * adjacent layout aligns with the cap and baseline of the text inside.
+   *
+   * This logic is a little fuzzy. Offsets of -0.25em and -0.5em work
+   * perfectly for a line height of 1.5, but they start to drift as it gets
+   * smaller or larger, so we further scale the value based on its difference
+   * from 1.5. Experiments suggest this produces reasonably accurate results.
+   */
+  const topOffset = `${lineHeight / -6.0 * (lineHeight / 1.5)}em`;
+  const bottomOffset = `${lineHeight / -3.0 * (lineHeight / 1.5)}em`;
```
zelitzsi: It'd be nice to have a page that tests these spacing values across a wide range of font sizes, perhaps on the typography page.

a-type: Would be a good expansion as we add text components. For now, the headings themselves are the only uses.

a-type: Yeah, I guess I'm saying this is good enough for now since the only components that currently do use this are the H elements you can see above.

PR #112, opened 2018-05-03
Fix spacing in various components

a-type:
## PR Checklist

## Breaking Changes

* Fixed incorrect padding on small accordion content

* Fixed bug with `spacing` prop in headings
* Extended `spacing` prop to allow specifying `vertical` or `horizontal`
* Fixed small accordion spacing
* Fixed button size
* Aligned button text vertically

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -5,8 +5,14 @@ import icons from '../Icon/icons';
 import SubmitButton from './SubmitButton';
 import get from 'extensions/themeGet';
 
+/**
+ * NOTE: all paddings in button are adjusted to accomodate
+ * border widths on both sides. If border widths are changed,
+ * also change padding to compensate!
+ */
+
 const Button = styled.button`
-  font-size: 0.8em;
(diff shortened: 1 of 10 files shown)


src/components/Button/Button.js
```diff
@@ -5,8 +5,17 @@ import icons from '../Icon/icons';
 import SubmitButton from './SubmitButton';
 import get from 'extensions/themeGet';
 
+/**
+ * NOTE: all paddings in button are adjusted to accomodate
+ * border widths on both sides. If border widths are changed,
+ * also change padding to compensate!
+ *
+ * Vertical paddings are also displaced to accomodate text cap and base lines.
+ * Basically, 1px is added to the top, 1px subtracted from the bottom.
```
zelitzsi: Doesn't `box-sizing` fix this issue?

a-type: Nope, since the height is determined by padding, not an explicit `height` property.

Can't set `height` explicitly since the text inside might need to wrap in some circumstances.

PR #113, opened 2018-05-03
Move to react-spring for accordions for native performance

a-type:
gif

(it's actually smoother IRL)

## PR Checklist

Check these before submitting your pull request:

- [x] Visual and behavioral components are separated
- [x] Exported components are documented with examples
- [x] Props have JSDoc comments
- [x] All relevant visual sub-components can be overridden via props

## Breaking Changes

* No changes in usage

## Changes to Existing Components

* Replaced the underlying animation tech behind the ExpandToggle component to use react-spring in `native` mode. Should be more performant!

PR #114, opened 2018-05-07
revert react-spring

zelitzsi:
## PR Checklist

Revert of #113 - it appears that `react-spring` is not ready for production projects yet, as we have encountered numerous problems with it. In particular, it has an issue that conflicts with `react-dnd`.

This commit was made using `git revert <hash>`.

### DIFF
--- a/src/behaviors/ExpandToggle/ExpandToggle.js
+++ b/src/behaviors/ExpandToggle/ExpandToggle.js
@@ -1,8 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import _ from 'lodash';
-import { Transition, Spring, animated } from 'react-spring';
-import FixAuto from './FixAuto';
+import { Collapse, UnmountClosed } from 'react-collapse';
 
 /**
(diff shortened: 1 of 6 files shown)


src/behaviors/ExpandToggle/ExpandToggle.js
```diff
@@ -164,7 +123,9 @@ class ExpandToggle extends React.Component {
         >
           {this.renderToggle()}
         </div>
-        {unmountClosed ? this.renderUnmountClosed() : this.renderMountClosed()}
+        <CollapseType isOpened={isExpanded} springConfig={springConfig}>
```
tbowenbw: should `springConfig` be renamed since we're no longer using react-spring?  Or is spring a more general term for collapsing behavior?

zelitzsi: Spring is a common way to describe a certain type of animation, named for basically the physics concept of springs that it follows. `react-collapse` uses some kind of spring animation in its internals, which I think is what we are configuring.

a-type:
Seems ready to merge to me.

PR #130, opened 2018-05-30
Fix react-router not performing exact matches

tato123:

Updates to the Anchor component based on the behavior of how react router performs matching. There are two cases in particular that can be problematic when using the `!!match` 

- Programmer intends for an exact match but the match object returns a null match [valid case, see react router documentation on null matches](https://reacttraining.com/react-router/web/api/match/null-matches)
- An exact match but the url contains special characters (e.g. `frame:foo` and `frame:bar` are evaluated as an exact match by react router because the colon is treated as a parameter, causing this to evaluate to true).

### DIFF
--- a/src/components/Anchor/Anchor.js
+++ b/src/components/Anchor/Anchor.js
@@ -208,6 +208,19 @@ class Anchor extends React.Component {
     }
   };
 
+  isActive(to, match, exact) {
+    let active;
+    if (!exact) {
+      // use the default behavior for non-exact matches
+      active = !!match;
+    } else {
+      // when we expect an exact match, dont allow react-router null
+      // value, additionally verify special chars are used
+      active = match == null ? false : match.url === to;
+    }
+    return active;
+  }
+
   render() {
     const {
       to,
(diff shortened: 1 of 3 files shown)


src/components/Anchor/Anchor.js
```diff
@@ -208,6 +208,19 @@ class Anchor extends React.Component {
...
 
+  isActive(to, match, exact) {
+    let active;
+    if (!exact) {
+      // use the default behavior for non-exact matches
+      active = !!match;
+    } else {
+      // when we expect an exact match, dont allow react-router null
+      // value, additionally verify special chars are used
+      active = match == null ? false : match.url === to;
```
a-type: Isn't there a `match.exact` field for this?

a-type: `isExact`, actually https://reacttraining.com/react-router/core/api/match

tato123: `isExact` value is incorrect as well, sample output shown below by adding a console log statement. `/frame:manageexternalnumbers` and `/frame:numberassignment` both have an `isExact` value of true even though the current url is `/frame:numberassignment`

```
[SCL] {path: "/frame:manageexternalnumbers", url: "/frame:numberassignment", isExact: true, params: {…}}
Anchor.js:180 
[SCL] {path: "/frame:numberassignment", url: "/frame:numberassignment", isExact: true, params: {…}}
```

PR #132, opened 2018-05-31
add spacing support to anchor

zelitzsi:
# Anchor Spacing
anchorspacing

## PR Checklist

* The `Anchor` component now supports a `spacing` prop to control spacing to it. I elected to have a default of zero spacing because anchors are commonly used within other elements (paragraphs, buttons, etc.)

Image comparisons are included below - they are very close (the new version ends up on full pixel values instead of decimal pixel values).

## Old Sizes
oldanchorsize
oldanchorfocusedsize

## New Sizes
newanchorsize
newanchorfocusedsize

### DIFF
--- a/src/components/Anchor/styles/ExternalIconAnchor.js
+++ b/src/components/Anchor/styles/ExternalIconAnchor.js
@@ -8,7 +8,6 @@ import { base, positive, negative, dark, inverted } from './sharedStyles';
 
 const iconStyles = css`
   text-decoration: none;
-  display: inline-block;
   text-transform: uppercase;
   font-weight: 700;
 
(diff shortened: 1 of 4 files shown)


src/components/Anchor/styles/ExternalIconAnchor.js
```diff
@@ -8,7 +8,6 @@ import { base, positive, negative, dark, inverted } from './sharedStyles';
 
 const iconStyles = css`
   text-decoration: none;
-  display: inline-block;
```
a-type: This seems like it has the possibility of breaking something

zelitzsi: I did a run through and realized I needed `auto` on the horizontal spacing by default. After that, everything seemed to work properly. I added visual comparisons as well to see what's going on and demonstrated the anchor's use within other text.

PR #133, opened 2018-06-01
Adding disable prop to Anchor

tbowenbw:
## PR Checklist

## Breaking Changes

* None

* Added a `disable` prop to Anchor, which works in much the same fashion as the `disable` prop on other form components (e.g. Button)

## New Visual Components

* none

## New Behavioral Components

* none


src/components/Anchor/Anchor.js
```diff
@@ -163,7 +169,9 @@ class Anchor extends React.Component {
   };
 
   handleClick = event => {
-    const { onClick, to } = this.props;
+    const { onClick, to, disabled } = this.props;
+    // Don't handle clicks if disabled
+    if (disabled) return;
```
tbowenbw: Does this need an `event.preventDefault`?  It seems to work from styleguide, but I'm not certain

zelitzsi: I'm not sure we need to do this at all if we apply `pointer-events: none` on the anchor when it's disabled.

src/components/Anchor/styles/sharedStyles.js
```diff
@@ -28,78 +30,83 @@ export const base = css`
...
-    height: 1px;
-    width: 100%;
-    position: absolute;
-    bottom: -0.1em;
-    left: 0;
-    transition: height 0.15s ease, width 0.15s ease, left 0.15s ease,
+  ${({ disabled, appearFocused }) =>
+    disabled
+      ? ''
+      : `
```
a-type: I'm not sure interpolations consistently work with a plain template string, I believe that's what the `css` export from `styled-components` is meant to provide.

src/components/Anchor/styles/sharedStyles.js
```diff
@@ -25,7 +25,7 @@ export const base = css`
   height: auto;
   margin: auto;
 
-  &:focus {
+  ${({ disabled }) => (disabled ? 'pointer-events: none' : '')} &:focus {
```
zelitzsi: Do we need a `;` between these two statements? Then they can be on separate lines.

a-type: Probably Prettier doing that. It's syntactically ok, but yeah, I wish they'd improve the parsing on CSS.

src/components/Anchor/Anchor.js
```diff
@@ -170,8 +170,6 @@ class Anchor extends React.Component {
 
   handleClick = event => {
     const { onClick, to, disabled } = this.props;
-    // Don't handle clicks if disabled
-    if (disabled) return;
```
a-type: To make disabled function correctly, I think we should go with your original proposal... return early while calling `event.preventDefault();`.

That makes it so that a user can replace the entire Anchor rendering component and still have disabled work.

PR #136, opened 2018-06-06
Spacing fixes

zelitzsi:
## PR Checklist

* Tweaked spacing between certain components (Priority Groups, Checkboxes, Toggles, Radios, etc.) to be smarter. Previously, we were doing a lot of checks for `:last-child`, but it fails if the component is not the last child. We now use the `+` operator instead to better describe the behavior and fix this issue.

checkboxspacing
togglespacing

* Tweaked the spacing values for `userSpacing` to be closer to design spec. I arrived at these numbers by looking at how tall our inputs are supposed to be in the design system and figuring out the baseline to cap line height from that. The sizes drift slightly as the fonts get bigger, but overall they are closer to spec.

fontsizes

* Fixed the spacing of the large size `Accordion` component

accordionspacing
accordionopen

* Fixed small amounts of sizing error on toggles and checkboxes. The bounding boxes for those elements should be correct.
* Adjusted inputs to be the correct 53px size.

## New Visual Components

* Added `TextDiv` component – this is just a `<P>` component except with a `<div>` instead of a `<p>`. The idea is that this component can wrap things that spacing doesn't work nicely for (i.e., anchors).


src/components/Checkbox/styles/CheckboxLabel.js
```diff
@@ -1,17 +1,18 @@
 import styled from 'styled-components';
 import get from 'extensions/themeGet';
-import Label from '../../Label';
 import Input from './CheckboxInput';
 import icons from 'components/Icon/icons';
 
 const SIZE = '30px';
 const CHECK_SIZE = '21px';
 
 // Builds off the base label
```
a-type: This is no longer true.

src/components/DragList/DragList.js
```diff
@@ -6,8 +6,8 @@ import DragListItem from './DragListItem';
 import DragContainer from 'components/DragContainer';
 
 const StyledDiv = styled.div`
-  & > * {
-    margin-bottom: ${get('spacing.medium')};
+  & > * + * {
```
a-type: so advanced

src/components/Input/styles/InputStyles.js
```diff
@@ -7,7 +7,7 @@ const InputStyles = styled.input`
   font-size: 14px;
   font-family: ${get('fonts.brand')};
   transition: all 0.2s ease;
-  padding: ${get('spacing.medium')};
+  padding: calc(${get('spacing.medium')} - ${get('thicknesses.normal')});
```
a-type: This seems a bit arbitrary, since the border thickness is `thicknesses.wide`.

src/components/TextDiv/TextDiv.js
```diff
@@ -0,0 +1,46 @@
...
+ * Text divs are vertically aligned based on the cap and
+ * base lines of the text. They support the `spacing`
+ * property, which allows a user to specify custom spacing
+ * around the component.
+ *
+ * **NOTE:** Many components provide the `spacing` property
+ * already. This component is provided as a catch-all
+ * solution.
+ */
+const TextDiv = styled(tag.div).attrs({
```
a-type: I don't think we should really encourage use of this externally at all. Semantically, if you mean to put a paragraph, you should be using `P`. Not that I don't think we're all aware of that practice, but I have noticed that bad practices tend to get adopted quickly with such a wide-spread resource as this library.

Basically I think a nice big `USE P INSTEAD IF YOU NEED A PARAGRAPH OF TEXT` would be a good addition, or even maybe not documenting this in the site.

zelitzsi: The reason I added this was because of the problems with anchors (specifically, `inline-block` not accepting negative margins). I can wrap it in a `<P>`, but it's not really a block of text.

src/components/Accordion/Accordion.js
```diff
@@ -105,6 +105,7 @@ class Accordion extends React.Component {
     Arrow: AccordionArrow,
     LabelText: AccordionLabelText,
     Content: AccordionContent,
+    spacing: 'lg',
```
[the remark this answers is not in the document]
zelitzsi: Are you talking about the spacing within the accordion around the content? I have to check if that's really the case for small accordions too.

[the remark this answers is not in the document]
a-type: I thought it was. At the very least, this should probably be `labelSpacing` to avoid confusion.

src/components/TextDiv/TextDiv.js
a-type: Yeah, it makes sense for internal use, I just worry about it becoming a temptation to lazy element selection if advertised as a library feature.

a-type: I guess you mean you want to use it externally, though. I really feel like there should be a way to more seamlessly fix the anchors. Like...

```
display: block;
margin: ${props => props.spacing};

p > &,
span > & {
  display: inline;
  margin: auto;
}
```

src/components/Accordion/Accordion.js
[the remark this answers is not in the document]
zelitzsi: Checked with UX, apparently it is supposed to be 30px padding inside the small accordion content. I can change the name though to indicate that it's the internal spacing.

src/components/TextDiv/TextDiv.js
zelitzsi: I can try that, though it might be confusing to. change the behavior based on the parent .

src/components/Input/styles/InputStyles.js
a-type: What I mean is, I'd be happier with just `1px`. I think using a named value implies more specific meaning, which can be confusing.

src/extensions/userSpacing.js
```diff
@@ -55,8 +55,8 @@ const userTextSpacingWithLineHeight = lineHeight => props => {
    * smaller or larger, so we further scale the value based on its difference
    * from 1.5. Experiments suggest this produces reasonably accurate results.
    */
-  const topOffset = `${lineHeight / -6.0 * (lineHeight / 1.5)}em`;
-  const bottomOffset = `${lineHeight / -3.0 * (lineHeight / 1.5)}em`;
+  const topOffset = `${lineHeight / -5.25 * (lineHeight / 1.5)}em`;
```
tato123: Is `-5.25 ` and `-3.5` (below) proportional to something? I'm nervous these values are going to be proportionate to  a particular width and font-size, meaning it will work only in certain situations (which might also mean that the values above worked for a particular page)

PR #137, opened 2018-06-08
Use field-day for Fields

a-type:
Pivots the Fields layout to use a library I've been working on which simplifies the overall structure / logic. All examples are still visually consistent with the exception of the breaking changes noted below.

## PR Checklist

## Breaking Changes

* Field.Group now has a 30px bottom margin (this is consistent with our spacing flow, and will actually help eliminate some customizations / Spacing component usage)
* Field help text now aligns against the field content directly, rather than being aligned as a separate row in a Field configuration
* Underlying HTML structure of Fields has been simplified. This may affect selectors or other obscure use cases

* Fields now uses the field-day library
* Fields support vertical alignment prop for content
* Fields support alignment props for label and helptext
* Fields now accepts id and className and style props
* You can now import Field directly and reference Field.Group, rather than the opposite


a-type:
I'm going to hold this off for a patch release after our next minor release.

PR #153, opened 2018-06-25
Use CSS vars to theme components

zelitzsi:
topbar

## PR Checklist

* We now use CSS vars to achieve theming. `themeGet` now effectively just converts the string to the CSS var version, so it's still useable but deprecated. Themes are applied on the root in `bootstrap()` with an option to apply a different theme programmatically via `BandwidthProvider` (in case you want to modify the theme on the fly). Alternatively, we could just apply the theme to `:root` in `BandwidthProvider`, but that breaks the parent-child relationship of the DOM
* The docs now include a theme switcher at the top of every page. The normal component controls have been removed
* The site now has a new logo look that is closer to dev.bandwidth.com

## New Visual Components

* Blue version of Bandwidth logo

## NOTE:
* I also changed all references of `themeGet` over to CSS vars; rather than clog up this review with that, I am submitting most of it as a separate PR. However, some of it snuck into this review. If we think that tool would be useful, I could include it, but didn't seem like a long term inclusion in the library
* [CSS vars have pretty good global coverage](https://caniuse.com/#feat=css-variables), but as with any new technology, we should consider if it is worth an inclusion or if we should have a fallback. It effectively knocks out IE support altogether.


tools/styleguide/StyleGuideRenderer.js
```diff
@@ -35,30 +40,53 @@ const Footer = styled.div`
...
-        </Footer>
-      </Content>
-      <Sidebar>
-        <Logo />
-        {toc}
-      </Sidebar>
-    </Root>
-  </BandwidthProvider>
-);
+// export const themeEvents = new EventEmitter();
```
a-type: Unused?

zelitzsi:
I think I will probably try aliasing it to `StyleRoot`, since that's closer to the intention of BandwidthThemeProvider's usage. I'd also be okay with a version bump, it's been deprecated for several months.

a-type:
Yeah, my philosophy is to not major bump until we need to, then sweep up all the deprecations at that time. ThemeProvider may have actually survived the last major bump accidentally... I forget when it was deprecated.

PR #156, opened 2018-06-26
Change isActive to be configurable

tato123:
## PR Checklist

- [*] Visual and behavioral components are separated
- [*] Exported components are documented with examples
- [*] Props have JSDoc comments
- [*] All relevant visual sub-components can be overridden via props

Modifies the `Anchor` component to allow overriding the `active` as either a `Boolean` value or a `function` that implements the previous signature. 

This is necessary since some applications may have complex navigation rules where simply checking whether a child exists within a parent does not always work. For example, when using a hash based router child routes may not "match" parent routes (e.g. `/#foo` and `/#bar` may have a parent child relationship as part of the `Navigation` components, but anchor will not show the parent route as active). 

This change allows the implementor to choose how a link is activated (or potentially activate multiple naivgation links)

### DIFF
--- a/src/components/Anchor/Anchor.js
+++ b/src/components/Anchor/Anchor.js
@@ -29,6 +29,19 @@ const inferType = children => {
   return 'wrap';
 };
 
+const matchReactRouter = (to, match, exact) => {
+  let active;
+  if (!exact) {
+    // use the default behavior for non-exact matches
+    active = !!match;
+  } else {
+    // when we expect an exact match, dont allow react-router null
+    // value, additionally verify special chars are used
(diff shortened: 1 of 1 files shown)


src/components/Anchor/Anchor.js
```diff
@@ -29,6 +29,19 @@ const inferType = children => {
   return 'wrap';
 };
 
+const matchReactRouter = (to, match, exact) => {
+  let active;
+  if (!exact) {
```
tbowenbw: This could probably be refactored to avoid the need for declaring a variable, but that's mostly a stylistic issue, which is always subjective.

PR #162, opened 2018-06-28
X modal

tbowenbw:
## PR Checklist

- [?] All relevant visual sub-components can be overridden via props

## Breaking Changes

* `onBlockerClicked` prop removed from `Modal` and replaced with `onClose` instead, which is fired when either the user clicks on the `ModalBlocker` or when the user clicks on the `Anchor` embedded in the default `ModalTitle`

* Adding an X to existing Modal

## New Visual Components

None
## New Behavioral Components

None


tbowenbw:
Unsure if inserting an `Anchor` into the `Title` lets the developer override all sub-components.  Perhaps `ModalTitle` should be a proper `React.Component` which itself uses the `H3` and `Anchor` so that the developer can override this with another component if desired.  Interested specifically in feedback around this topic.

src/components/Modal/Modal.js
```diff
@@ -18,9 +19,9 @@ class Modal extends React.Component {
      */
     title: PropTypes.string,
     /**
-     * Handles click events on the backdrop behind the modal. Use this to close it if desired.
+     * Handles click events on the backdrop behind the modal or on the X in the title bar.
      */
-    onBlockerClicked: PropTypes.func,
+    onClose: PropTypes.func,
```
zelitzsi: We should mark `onClose` as required.

a-type: Some modals are closed by internal button controls, which would force the user to pass a no-op to the modal component itself.

src/components/Modal/Modal.js
```diff
@@ -77,25 +78,34 @@ class Modal extends React.Component {
     event.stopPropagation();
   };
 
+  renderTitle = () => {
+    const { title, onClose, Title } = this.props;
+    if (!title) return null;
+    return (
+      <Title>
+        {title}
+        <Anchor icon="delete2" onClick={onClose} />
```
a-type: Adding a truthiness check for `onClose` to conditionally render the anchor would let us not render a close button that doesn't do anything.

src/components/Modal/Modal.js
```diff
@@ -77,25 +78,34 @@ class Modal extends React.Component {
...
       Blocker,
       Content,
       Window,
       ActionContent,
-      Title,
     } = this.props;
 
     return (
-      <Blocker onClick={onBlockerClicked}>
+      <Blocker onClick={onClose}>
```
a-type: Please fallback to `onBlockerClicked` to keep compatibility.

src/components/Modal/Modal.js
a-type: Also, to adhere to the 'overrideable props' rule, you'll want to make `CloseButton` a prop for this component and use it here, and provide the SCL anchor as a default prop.

src/components/Modal/styles/ModalTitle.js
```diff
@@ -1,6 +1,12 @@
 import SectionTitle from 'components/SectionTitle';
 import styled from 'styled-components';
+import Anchor from 'components/Anchor';
 
 export default styled(SectionTitle)`
   padding: 0.95em 1em 0.95em 1.5em;
+
+  & > a {
```
a-type: A stronger selector would be to pass a class name to the close button anchor and reference it here instead. This works hand-in-hand with the overrideable prop suggestion.

PR #169, opened 2018-07-19
Added TimePicker & DateTimePicker

zelitzsi:
## PR Checklist

* Improvements to`DatePicker` and `DatePicker.Range`:
    * Both now automatically handle focus state (can be overridden)
    * Both now have disabled and invalid styling
* Added `user-select: none` to `RadioButton` to make it easier to click (previously it was easy to select the text inside the button instead)
* `RadioButton` now uses `display: inline-block` so that they look right even without a parent wrapper

## New Visual Components

* New `TimePicker`, `DateTimePicker`, and `DateTimePicker.Range`. The latter two integrate the `TimePicker` into the preexisting `DatePicker` and `DatePicker.Range` components.
* All of these new components support both controlled/uncontrolled states and use the standard `value` and `onChange` props
* All of these new components support utc format

datetimerangepicker


src/components/DatePicker/DatePicker.js
```diff
@@ -9,35 +9,82 @@ import Icon from '../Icon';
 import DatePickerWrapper from './styles/DatePickerWrapper';
 
 import DateRangePicker from './DateRangePicker';
+import moment from 'moment';
+moment.updateLocale('en', {
+  weekdaysMin: ['S', 'M', 'T', 'W', 'T', 'F', 'S'],
```
a-type: Is this a global change? Can we contain it to a copy of moment used just for this component?

src/components/TimePicker/TimePicker.js
```diff
@@ -0,0 +1,94 @@
...
+const RadioButton = RadioGroup.Button;
+import moment from 'moment';
+import Container from './styles/Container';
+import RadioContainer from './styles/RadioContainer';
+import StyledInput from './styles/StyledInput';
+
+class TimePicker extends React.PureComponent {
+  static propTypes = {
+    /**
+     * TODO: Rewrite this description
```
a-type: 🙈

src/components/TimePicker/TimePicker.js
```diff
@@ -0,0 +1,94 @@
...
+  state = {
+    internalValue: undefined,
+  };
+
+  handleOnChange = ({ value, amSelected }) => {
+    if (!value) {
+      return;
+    }
+    const timeValue = moment(value, this.props.timeFormat);
+    const hours = amSelected === timeValue.hours() >= 12 ? 12 : 0;
```
a-type: This statement is pretty confusing to read. `amSelected equals timevalue hours greater than or equal to twelve, then twelve else zero`? Seems like there's some JS order-of-operations needed to parse... would be preferable to simplify and break down.

zelitzsi:
I was mostly curious to see how well they worked, plus I think they make sense given how I implemented the functionality.

src/components/DatePicker/DatePicker.js
zelitzsi: It is, and it also only applies for English.

zelitzsi: It looks like we would have to hack around this limitation in a similarly ugly way. It might be better to leave this hacky code in here for now.

PR #176, opened 2018-08-02
Move to a monorepo structure, connect layout and scl libraries

a-type:
I think it's reasonable not to review this based on code changes. Go to https://github.com/Bandwidth/shared-components/tree/monorepo to see the structure of the repo after the changes.

I have moved all packages under `/packages` and added the `layouts` project to this repo. I've also set up the `lerna` tool in an attempt to keep the projects synchronized. I'm staging all this under a release candidate for `5.0.0`.

This prevents us from having to go through hurdles of repo management as we continue to decompose the role of the SCL into smaller pieces.


zelitzsi:
I had to install babel globally in order to build because it is used in `layouts/package.json`.

a-type:
@zelitzsi probably this problem: https://github.com/reactjs/react-docgen/issues/203

I'll look into it.

PR #184, opened 2018-08-15
BWDB-3718 SCL - Header bar

dlisnichenko-cv:
## PR Checklist

## New Visual Components
This pull request adds a header bar that may be reused in other places across Bandwidths' portals. 

There should be a new horizontal 'header bar' for use between the navigation menu and the main content on any given page. 

The header bar should be 50px tall
The header bar should have a bottom border.
style: 1px solid #c2c2c2;
create


dlisnichenko-cv:
Updated patch according to a new project structure

packages/shared-components/src/components/HeaderBar/HeaderBar.js
```diff
@@ -0,0 +1,11 @@
+import styled from 'styled-components';
+
+const HeaderBar = styled.div.withConfig({ displayName: 'HeaderBar' })`
+  height: 50px;
+  border-bottom: 1px solid #c2c2c2;
```
zelitzsi: Please use themeGet to use the theme value for this color and line thickness.

dlisnichenko-cv:
Fixed review comments 
scr1

packages/shared-components/src/components/HeaderBar/HeaderBar.js
```diff
@@ -1,8 +1,11 @@
 import styled from 'styled-components';
+import irisTheme from 'theme/irisTheme';
 
 const HeaderBar = styled.div.withConfig({ displayName: 'HeaderBar' })`
   height: 50px;
-  border-bottom: 1px solid #c2c2c2;
+  display: flex;
+  align-items: center;
+  border-bottom: ${irisTheme.thicknesses.normal} solid ${irisTheme.colors.gray.border};
```
zelitzsi: Use `themeGet()`; it will adapt to the current theme rather than assuming the current theme is IRIS (a relatively safe assumption for now, but not future proof).

dlisnichenko-cv: ThemeGet is marked as deprecated. Ok, will switch to it

dlisnichenko-cv:
Switched to themeGet and added padding
scr1

packages/shared-components/src/components/HeaderBar/HeaderBar.js
zelitzsi: Yeah. For some history, we deprecated themeGet because we wanted to switch to using css vars. However, we are still determining how we are supporting IE, so we should continue to use it for the meantime. The alternative would be to use css vars in styled components directly.

packages/shared-components/src/components/HeaderBar/HeaderBar.js
dlisnichenko-cv: ok, done

PR #186, opened 2018-08-16
UI-197 fix select

lunxiao:

Fixed the Select component so that empty placeholder is displayed correctly.

### DIFF
--- a/packages/shared-components/src/components/Select/Select.js
+++ b/packages/shared-components/src/components/Select/Select.js
@@ -227,7 +227,7 @@ class Select extends React.PureComponent {
       Wrapper,
     } = this.props;
     const combinedLoading = loading || isLoading;
-    const combinedPlaceholder = placeholder || noneText;
+    const combinedPlaceholder = (noneText !== undefined) ? noneText : placeholder;
 
     return (
       <Wrapper invalid={invalid}>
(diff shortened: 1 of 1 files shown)


packages/shared-components/src/components/Select/Select.js
```diff
@@ -227,7 +227,7 @@ class Select extends React.PureComponent {
       Wrapper,
     } = this.props;
     const combinedLoading = loading || isLoading;
-    const combinedPlaceholder = placeholder || noneText;
+    const combinedPlaceholder = (noneText !== undefined) ? noneText : placeholder;
```
Aubron: noneText is marked as deprecated in the documentation. Wouldn't a more consistent solution be to have the placeholder not rendered if `placeholder === null`?

lunxiao: placeholder has a default value and will never be null.

Aubron: That just feels like the scope of the issue we're wanting to solve here. We want the component to longer have a default for placeholder, but instead allow an empty placeholder. Although maybe the usage of this component in the existing projects would cause issues with that manner of doing it (places where we're relying on the placeholder to give us None:).

From solely a component architecture perspective though, we would typically want to limit our number of custom props and emulate html props, making placeholder a good candidate for this change. Emulating the html input implementation of placeholder would make this component easier for a newcomer to use without documentation, and would just require not passing a default to it.

At the very least, if we do end up re-implementing noneText (which probably should be renamed to something like a bool hasPlaceholder, we need to change it's Proptypes documentation on L156

PR #196, opened 2018-08-30
Color for required mark now is coming from default colors

denysYefremov:

* Changed asterisk color from static to the color from the default set of colors


denysYefremov:
Are there extra steps I need to do before merging this commit?

zelitzsi:
No, you should be good.

PR #199, opened 2018-09-04
UI-227 update FileLoader per UX requests

lunxiao:

FileLoader placeholder text has two sections as shown below:

image

### DIFF
--- a/packages/shared-components/src/components/FileLoader/FileLoader.js
+++ b/packages/shared-components/src/components/FileLoader/FileLoader.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Dropzone from 'react-dropzone';
 import Icon from '../Icon';
+import Anchor from '../Anchor';
 import FileLoaderDropArea from './styles/FileLoaderDropArea';
 import FileLoaderPreview from './styles/FileLoaderPreview';
 
@@ -59,7 +60,7 @@ export default class FileLoader extends React.Component {
       return (
         <DropArea>
           <Icon name="file" />
-          <Preview><span>DROP A FILE HERE, OR </span><span>CLICK TO BROWSE</span></Preview>
+          <Preview>DROP A FILE HERE, OR&nbsp;<Anchor>CLICK TO BROWSE</Anchor></Preview>
         </DropArea>
       );
     }
--- a/packages/shared-components/src/components/FileLoader/FileLoader.js
+++ b/packages/shared-components/src/components/FileLoader/FileLoader.js
@@ -59,7 +59,7 @@ export default class FileLoader extends React.Component {
       return (
         <DropArea>
           <Icon name="file" />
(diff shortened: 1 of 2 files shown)


packages/shared-components/src/components/FileLoader/styles/FileLoaderPreview.js
```diff
@@ -1,5 +1,13 @@
 import styled from 'styled-components';
+import get from 'extensions/themeGet';
 
 export default styled.div`
-  margin: auto auto auto 16px;
+  margin: auto 4px auto 16px;
+  & > * {
+    &:last-child {
+      margin: auto auto auto 2px;
```
zelitzsi: Can't we just use a `&nbsp;`?

packages/shared-components/src/components/FileLoader/styles/FileLoaderPreview.js
```diff
@@ -1,5 +1,13 @@
...
+      color: ${get('colors.primary.alternate')};
+      text-decoration: underline;
```
zelitzsi: This seems off. I would expect the `CLICK TO BROWSE` item to be styled the same as our anchors.

packages/shared-components/src/components/FileLoader/FileLoader.js
```diff
@@ -59,7 +59,7 @@ export default class FileLoader extends React.Component {
       return (
         <DropArea>
           <Icon name="file" />
-          <Preview>Drop a file here, or click to browse.</Preview>
+          <Preview><span>DROP A FILE HERE, OR </span><span>CLICK TO BROWSE</span></Preview>
```
zelitzsi: Does anything special happen if the user clicks on the `CLICK TO BROWSE` part vs the `DROP A FILE HERE` part?

zelitzsi: The `DROP A FILE HERE` text looks like it's supposed to be a lighter color than standard text.

zelitzsi: The mockup doesn't have the paper icon on the left side of the box.

packages/shared-components/src/components/FileLoader/styles/FileLoaderPreview.js
zelitzsi: The mockup seems to have bold text here as well?

a-type: Seems like we should just use an Anchor component here with no click behavior (since the whole input is clickable).

packages/shared-components/src/components/FileLoader/FileLoader.js
lunxiao: no, according to Shaun the behavior is the same. All he wants is the underlined and different color of text.

lunxiao:
Shaun said leave the icon there. The color is ok too.

PR #200, opened 2018-09-04
radioButton fix styles

lunxiao:
UX requirements:

"there are two sizes. the normal (larger) size should be 53px tall, min-width of 53px, fixed left/right padding of 15px, base font size (14px). the smaller size should be 30px tall, min-width of 30px, fixed left/right padding of 10px, font size 0.8em (11.2px)
all instances should have horizontally and vertically centered text in the button and each button is sized according to its content, not any external container that link implies that the SCL doesn’t match these requirements so we should update it"

image

### DIFF
--- a/packages/shared-components/src/components/RadioGroup/styles/RadioGroupButtonLabel.js
+++ b/packages/shared-components/src/components/RadioGroup/styles/RadioGroupButtonLabel.js
@@ -6,7 +6,7 @@ const RadioGroupButtonLabel = styled.label`
   border-width: ${get('thicknesses.thick')};
   border-style: solid;
   border-color: ${get('colors.primary.dark')};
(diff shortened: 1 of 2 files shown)


packages/shared-components/src/components/RadioGroup/styles/RadioGroupButtonLabel.js
```diff
@@ -6,7 +6,7 @@ const RadioGroupButtonLabel = styled.label`
   border-width: ${get('thicknesses.thick')};
   border-style: solid;
   border-color: ${get('colors.primary.dark')};
-  padding: 1em;
+  padding: 1em 15px;
```
zelitzsi: Use the 15px value from the theme?

PR #204, opened 2018-09-07
UI-236 fix seachBox for firefox

lunxiao:

SearchBox search button does not work on Firefox due to the fact that button (hidden) width is 0. I had to set a width to it and then adjust its position to the :before element.

image

### DIFF
--- a/packages/shared-components/src/components/SearchBox/styles/SearchBoxSearchButton.js
+++ b/packages/shared-components/src/components/SearchBox/styles/SearchBoxSearchButton.js
@@ -8,18 +8,21 @@ const INACTIVE_SIZE = '2em';
 export default styled.button`
   position: relative;
   display: block;
-  width: 0;
+  width: ${ACTIVE_SIZE};
+  height: ${ACTIVE_SIZE};
+  line-height: ${ACTIVE_SIZE};
   overflow: visible;
   background: none;
   border: none;
   padding: 0;
   font-size: 1.5em;
   outline: none;
+  margin-left: -${ACTIVE_SIZE} !important;
 
   &::before {
     position: absolute;
     top: -0.125em;
-    left: calc(-${ACTIVE_SIZE} / 2);
+    left: 15px;
     font-family: Bandwidth;
     content: "${icons('search')}";
     width: ${ACTIVE_SIZE};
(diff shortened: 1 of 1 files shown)


packages/shared-components/src/components/SearchBox/styles/SearchBoxSearchButton.js
```diff
@@ -8,18 +8,21 @@ const INACTIVE_SIZE = '2em';
 export default styled.button`
   position: relative;
   display: block;
-  width: 0;
+  width: ${ACTIVE_SIZE};
+  height: ${ACTIVE_SIZE};
+  line-height: ${ACTIVE_SIZE};
```
zelitzsi: Do we need to change the `line-height`?

packages/shared-components/src/components/SearchBox/styles/SearchBoxSearchButton.js
```diff
@@ -39,13 +42,18 @@ export default styled.button`
...
+
   &:disabled::before {
+    position: absolute;
     width: ${INACTIVE_SIZE};
     height: ${INACTIVE_SIZE};
     top: 0.25em;
     line-height: ${INACTIVE_SIZE};
     font-size: 1em;
-    left: calc(-${INACTIVE_SIZE} - ${get('spacing.small')});
+    left: 10px;
```
zelitzsi: It's confusing how this is 10px and the other one is 15px. Does the button end up moving when it changes from disabled to enabled or vice versa?

packages/shared-components/src/components/SearchBox/styles/SearchBoxSearchButton.js
lunxiao: probably not. I just made it the same as the before element.

packages/shared-components/src/components/SearchBox/styles/SearchBoxSearchButton.js
lunxiao: the 10px makes the same space between the button (disabled) to the right border of the input. The 15px adjusts the position the big blue button has. 15px lines it right inside of the right border. The origin impl makes the blue button half outside of the input which looks strange to me.

PR #205, opened 2018-09-12
Timescrubber

a-type:
gif

## PR Checklist

## New Visual Components

* Various components used to construct the Scrubber

## New Behavioral Components

* A TimeScrubber which a user can use to select a time in 15 min increments


packages/shared-components/src/components/TimeScrubber/TimeScrubber.js
```diff
@@ -0,0 +1,375 @@
...
+    }
+    this.setState({ internalValue: newValue });
+    if (onChange) {
+      onChange(newValue);
+    }
+  };
+  incrementTime = () => this.stepTime(15);
+  decrementTime = () => this.stepTime(-15);
+  canIncrementTime = () =>
+    toHours(this.state.internalValue) < this.props.allowedRange[1];
```
zelitzsi: This assumes that the allowedRange provided is always even with the step value, which isn't the safest of assumptions.

packages/shared-components/src/components/TimeScrubber/constants.js
```diff
@@ -0,0 +1 @@
+export const BORDER_WIDTH = 2;
```
zelitzsi: Is this being used?

PR #209, opened 2018-09-26
Updated helpCallout to work with the whole text as the link which tri…

denysYefremov:
…ggers callout to show

## PR Checklist

Help callout was updated to allow user use text as the trigger for showing callout (it wraps by link). Container for the children and icon was removed because there is no need in it. The text will be always wrapped by some container, so it will keep text + icon together. But if there is an issue with this we can use `<div style={{ display: 'inline-block' }}>...</div>` because `flex` move all data to the next line


packages/shared-components/src/components/HelpCallout/HelpCallout.js
```diff
@@ -75,10 +107,7 @@ class HelpCallout extends React.Component {
 
     return (
       <Manager>
-        <div style={{ display: 'flex' }}>
-          {children}
```
zelitzsi: I actually wrote the children outside the reference here for a reason; the callout should be centered on the icon and not the text when you have the icon + text format.

zelitzsi:
@a-type We could just have it not include text, but it's really annoying to use then in the common use case where you have it paired with text, since you need to add a `<React.Fragment>` and `<Label>` wrapper around your text. Since we know it's the most common use case, we should make the component dirt simple to use in that case.

packages/shared-components/src/components/HelpCallout/HelpCallout.js
denysYefremov: But it centered on the icon, not the whole text, this is why you have `= ({ ref }) =>` in the `<Reference />`

zelitzsi: Oh right. Ignore me... 🐵

denysYefremov:
What do you think if we will use:

* Label + hoverable (?) - is there a reason to cover this case, it seems like we teach our component do more then 1 thing? The HelpCallout means that you use something that you will hover and it will show you popup with data (the same as title works on the elements) but not render some text just to align text with hovarable 
* hovarable Label + (?) - `<HelpCallout content="text">Label</HelpCallout>`
* Just (?) - `<HelpCallout content="text" />` 
* Miscellaneous text with a callout - `<HelpCallout withoutIcon content="text">Label</HelpCallout>` 

I like Grant idea to use children as the part of the trigger for the callout because we will have a huge flexability in this case. By use `children` as a part of the trigger, I mean not to wrap it with `Anchor` but use what ever you will provide inside

zelitzsi:
@denysYefremov I like the usage for the first and second examples. The fourth example has been a pattern we want to support, but the usage is a little bizarre. Perhaps we could natively support making it styled like an anchor?

a-type:
@zelitzsi 	I appreciate that we want to cater to common use cases. However, I think it makes the component usage kind of confusing when we start to support too many use cases which conflict semantically. To me, having the children be the hoverable element is the most intuitive usage.

As for the common case of appending a (?) to a Field label, I'd suggest we extend Field with a `helpCallout` prop which automatically appends a HelpCallout to the label contents with the provided content as the HelpCallout's `content` value. This keeps the `label` prop simple (just a string still) and makes the help callout usage even simpler (no need to even import the HelpCallout component).

zelitzsi:
@a-type In that case, we have almost the same usage as the `<Callout>` component, except with different behavior if no children are provided. I'm okay with that as a solution, thought it makes the anchor version slightly awkward.

@denysYefremov I like it, but with regards to the above comments, maybe we should change the usage to ```js
<Callout.Link>Some text<Icon name="help/></Callout.Link>
```

**Comment by denysYefremov** [comment]
I was thinking to improve callout and use it almost the same way as helpCallout, but we have usage of callout somewhere in the Janus, so if I will change the callout itself, can I replace all the callout in Janus after without task?

**Comment by denysYefremov** [comment]
Lets summarize all thing I need to do here:

1. Remove HelpCallout.Link;
2. Improve Callout
3. Add Callout.Link;
4. Extend Field to support HelpCallout;

Am I right?

**Comment by a-type** [comment]
I think maybe we should just be replacing Callout with HelpCallout? And use the `children` as the hover element. If there's no `children`, create a link with the (?) symbol instead. Then add `helpCallout` to Field.

File: packages/shared-components/src/components/Callout/Callout.js [hunk 2]
```diff
@@ -55,19 +68,55 @@ class Callout extends React.Component {
     clearTimeout(this._timer);
   };
 
+  renderReferenceContent = ({ ref }) => (
+    <div
```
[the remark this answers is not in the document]
denysYefremov: Not sure that I follow what should I do here. Did you mean that I should spread remaining props and pass them to the div, as the container?

PR #212, opened 2018-09-28
Clean up various stale code at Input component

dlisnichenko-cv:
## PR Checklist

## Breaking Changes

* Should not introduce any breaking changes. Only codestyle
Please see my comments below to understand why it is changed

### DIFF
--- a/packages/shared-components/src/components/Input/Input.js
+++ b/packages/shared-components/src/components/Input/Input.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { defaultProps } from 'recompose';
-import styled, { css } from 'styled-components';
 import Anchor from '../Anchor';
 import InputStyles from './styles/InputStyles';
 import InputRevealPasswordWrapper from './styles/InputRevealPasswordWrapper';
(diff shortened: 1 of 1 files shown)


packages/shared-components/src/components/Input/Input.js
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { defaultProps } from 'recompose';
-import styled, { css } from 'styled-components';
```
dlisnichenko-cv: not used

packages/shared-components/src/components/Input/Input.js
```diff
@@ -181,9 +180,8 @@ class Input extends React.PureComponent {
   };
 
   renderPasswordField = () => {
-    const { RevealPasswordWrapper } = this.props;
-    const toggleState = () =>
-      this.state._type === 'password' ? 'text' : 'password';
```
dlisnichenko-cv: already passed to function as arg

packages/shared-components/src/components/Input/Input.js
```diff
@@ -257,7 +254,6 @@ class Input extends React.PureComponent {
         readonly={readonly}
         placeholder={placeholder}
         innerRef={inputRef}
-        onBlur={onBlur}
```
dlisnichenko-cv: already passed above (`onBlur={this.onBlur}`)

PR #222, opened 2018-10-15
Skeletons support

zelitzsi:
loading

## PR Checklist

- [ ] ~Props have JSDoc comments~ Made an attempt here, but we have a bug where styled components don't display prop types

* Added `.Skeleton` to several components that seemed to really need a skeleton version

## New Visual Components

* `<Skeleton />` - Basic building block of skeleton components
* `<Skeleton.Group />` - Helper that defines a group of skeletons that are animated together. Can also easily repeat skeletons multiple times.


zelitzsi:
@a-type The original idea was that the component parts would be generally available and that consumers of the library could compose them to form other components. Even internally, we would compose them to form the major components. I can move the components that are specific to the right directories though.

a-type:
Atomic design may provide some guidance here. `Blob` is definitely an atom, for instance. But for things like `BlobParagraph`, it's really 1:1 analogous with `P`, and could be defined within the domain of the `P` code.

This has ended up working really well for me in the past. For instance, if you have a part of the UI called `Foo` that uses `P` and you want to create a loading state for it, you'll already have `import { P } from '@bandwidth/shared-components';`, so now it's dead simple to compose `P.Skeleton` into `Foo.Skeleton`...

```js
import React from 'react';
import { P, H1, Input } from '@bandwidth/shared-components';

const Foo = ({ title, content, value, onChange }) => (
  <div>
    <H1>{title}</H1>
    <P>{content}</P>
    <Input value={value} onChange={onChange} />
  </div>
);

Foo.Skeleton = () => (
  <div>
    <H1.Skeleton />
    <P.Skeleton />
    <Input.Skeleton />
  </div>
);

export default Foo;
```

Usage is easy and intuitive, and there's never any need to directly import a Blob-like component from the library itself.

src/components/Table/SkeletonTable.js
```diff
@@ -0,0 +1,71 @@
...
+  <SkeletonTableContainer
+    disableAnimation={disableAnimation}
+    width={width}
+    marginBetween={marginBetween}
+  >
+    <SkeletonTableHeader
+      disableAnimation={disableAnimation}
+      height={rowHeight}
+    />
+    {Array.from(Array(rows).keys()).map(rowIdx => (
```
denysYefremov: it seems like we can move this "array generator" to some utils and name it like: `generateArray(size)`, or `createArray(size)`

tbowenbw: This might be more idiomatic as: 
```
rows.map((row, rowIdx) => ( ... ));
```

zelitzsi: `rows` is a number though.

tbowenbw: ah, didn't notice that.  The chain of `Array.from(Array(rows).keys())` is pretty difficult to read.  I guess that means to use index parameter it'd have to be `Array(rows).map((_, rowIdx) => ( ... ))` which is still kinda difficult to read.

PR #225, opened 2018-10-18
Changed Modal to align with new styles (WIP)

denysYefremov:
Removed unused library from ScrollShadow;

What was implemented/already was in this component:

- Modals 150px from the top of the screen
- Modal height determined by the content
- Scrolling occur as needed based on the users viewport
- The buttons separated from the content by a dividing line that's 30px from the last bit of content
- The buttons 15px from the dividing line and from the bottom edge of the modal
- The bottom area with the action button should become fixed
- Modal buttons are right aligned
- Primary action is farthest to the right
- Modal height should never go beyond 150px from the bottom of the viewport
- If modal has a cancel/close secondary button, we do not show an ‘x’ in the top right corner

What was skipped:

- Modals use primary and secondary buttons (we rely on the user to chose right button)

## PR Checklist

Changed styles of Modal component regarding requirements


a-type:
I think we can rely on users to only use primary / secondary buttons, no enforcement needed.

I do think a layout component to do the right alignment and proper spacing would be beneficial. Would `Horizontal` maybe be helpful here?

denysYefremov:
From the new changes:

I've added 2 flags:
`withCancelButton` and `withCloseButton`, the only difference in the text they provide to button and both of them trigger `onClose` function

The reason is that we need to remove `X` icon in case if `cancel` or `close` button was provided, so I should have been handled this

Also, I've changed the `CloseButton` to `CloseIcon` because I think this the name this prop should have, most of the time we would render icon in the right corner, not button and from `onClose={onClose}` to `onClick={onClose}`, because I don't think that this is clear for developer, that he should provide `onClose` prop in case if he would use custom `CloseIcon`

packages/shared-components/src/components/Modal/Modal.js
```diff
@@ -93,10 +122,47 @@ class Modal extends React.Component {
...
+      buttonSpace,
+      withCloseButton,
+      withCancelButton,
+      onClose,
+    } = this.props;
+
+    const isCloseCancelButton =
+      (withCloseButton || withCancelButton) && onClose != null;
+
+    const CloseButton = isCloseCancelButton ? (
```
zelitzsi:
```suggestion
    const closeButton = isCloseCancelButton ? (
```

packages/shared-components/src/components/Modal/Modal.js
```diff
@@ -79,7 +104,11 @@ class Modal extends React.Component {
     Content: ModalContent,
     Title: ModalTitle,
     Window: ModalWindow,
-    CloseButton: ModalCloseButton,
+    CloseIcon: ModalCloseIcon,
+    alignButtons: 'right',
+    buttonSpace: 'sm',
```
zelitzsi: Suggestion: `spaceBetweenButtons`?

packages/shared-components/src/components/Modal/Modal.js
```diff
@@ -63,9 +61,36 @@ class Modal extends React.Component {
...
+    ]),
+    /**
+     * Render Cancel button which on click will call onClose function
+     */
+    withCancelButton: PropTypes.bool,
+    /**
+     * Render Close button which on click will call onClose function
      */
-    CloseButton: PropTypes.func,
+    withCloseButton: PropTypes.bool,
```
zelitzsi: I'm trying to figure out if these two buttons are better served as something provided by the consumer of this library (as they are now) or automatically like this. I think maybe these shouldn't be included.

packages/shared-components/src/components/Modal/Modal.js
```diff
@@ -93,10 +111,38 @@ class Modal extends React.Component {
...
+      actionContent,
+      ActionContent,
+      alignButtons,
+      spaceBetweenButtons,
+    } = this.props;
+
+    const isCloseCancelButton =
+      (withCloseButton || withCancelButton) && onClose != null;
+
+    const closeButton = isCloseCancelButton ? (
```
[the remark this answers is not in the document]
denysYefremov: Oh, sorry, its probably happens after merge =\

PR #229, opened 2018-10-23
SCL 6.0: merged libraries, rollup, cleaning

a-type:
See the [changelog](https://github.com/Bandwidth/shared-components/blob/96dc9471e8a5998432131c37fd8738f473aa186a/CHANGELOG.md) for a high-level review of what I've done so far.

Started working on merging the `layouts` and `shared-components` libraries, using Rollup to bundle them, and fixing any issues along the way.

There are still some broken parts. Also, I removed the auto-root-level-index stuff we previously used. It worked well enough, but hasn't scaled very well with the library or encouraged thoughtful usage patterns in my opinion.

I'm putting this out there as a trial balloon. There are some more things I'd like to look at for a vNext...

- [x] Merge layouts and components
- [x] Use Rollup
- [x] Remove raw `lodash` and use sub-packages instead
- [x] Remove deprecated components like `Flow`
- [x] Upgrade to `styled-components` v4
- [x] Stop importing font files inside the lib, instruct users to reference and include them as they see fit instead (needed for non-webpack compatibility)
- [x] Standardize all "grouped" components - you should import `X` and reference `X.Group`, not the other way around (lookin' at you, RadioGroup)
- [x] Simplify `Anchor`, remove dependency on React Router
- [x] Add indexes to all the `styles` folders to make importing their contents easier
- [x] Prevent library from emitting any errors during normal usage (fix proptypes, unknown attributes, etc)
- [x] Forward all "rest props" to underlying components
- [x] Move docs build to a `gh-pages` branch and automate it somehow to avoid docs changes polluting diffs and PRs
- [ ] ~~Standardize our animation techniques across the library~~ Save this for later.


docs/layouts/src/demos/SplitContent/Docs.js
```diff
@@ -8,7 +8,7 @@ import {
   Code,
   Spacing,
   Accordion,
-} from '@bandwidth/shared-components';
+} from '../../../../../src';
```
zelitzsi: Use absolute path

docs/layouts/src/demos/SplitContent/Left.js
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
-import { SplitContentLayout } from '../../../../src';
+import { SplitContentLayout, H2, Button, Anchor } from '../../../../../src';
```
zelitzsi: Use absolute path

docs/layouts/src/Home.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { H1, P, Spacing, Code, Anchor } from '@bandwidth/shared-components';
+import { H1, P, Spacing, Code, Anchor } from '../../../src';
```
zelitzsi: Use absolute path

docs/layouts/src/Navigation.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Navigation, Anchor } from '@bandwidth/shared-components';
+import { Navigation, Anchor } from '../../../src';
```
zelitzsi: Use absolute path

zelitzsi:
I will say that I think we should aim to fix the proptypes in a separate changeset from this one.

docs/layouts/src/demos/SplitContent/Docs.js
a-type: I don't think it would work in this context. I could maybe look into setting up a webpack alias.

a-type:
@zelitzsi most of the errors I've seen from SCL have been from Anchor. I think reworking Anchor (which unfortunately needs to happen, since styled-components v4 broke the very fragile glue holding it together) will put us close to getting them all worked out.

denysYefremov:
I agree with @zelitzsi we should use an absolute path. It is incredibly difficult, to maintain a relative path like this

src/components/Input/Input.js
```diff
@@ -0,0 +1,296 @@
...
+    placeholder: '',
+    disableShowPassword: false,
+    Styles: styles.InputStyles,
+    InlineContentWrapper: styles.InlineContentWrapper,
+    inlineContent: null,
+    maxLength: null,
+    min: null,
+    max: null,
+    autoComplete: true,
+    appearFocused: false,
```
denysYefremov: I think you missed to use this prop somewhere

Aubron:
I already submitted an approval for this, but just wanted to say I've been messing with this for some docs experiments, and it all seems good to me 👍

PR #230, opened 2018-10-23
components that need to generate their own ids will maintain them acr…

zelitzsi:
…oss renders

## PR Checklist

* Certain components need to have ids that match up between two DOM elements (using the html `for` attribute). These ids were being generated in the render function, which meant that they produced new DOM elements if they were rendered. Now, they will maintain the same id.

### DIFF
--- a/packages/shared-components/src/components/Checkbox/Checkbox.js
+++ b/packages/shared-components/src/components/Checkbox/Checkbox.js
@@ -11,46 +11,46 @@ export default class Checkbox extends React.PureComponent {
     /**
      * Adds a class name to the input element.
      */
-    className: PropTypes.string
+    className: PropTypes.string,
     /**
      * Adds an id to the input element.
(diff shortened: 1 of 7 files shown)


packages/shared-components/src/components/Checkbox/Checkbox.js
```diff
@@ -72,6 +72,8 @@ export default class Checkbox extends React.PureComponent {
     Label: CheckboxLabel.Small,
   })(Checkbox);
 
+  defaultId = generateId('checkbox');
```
denysYefremov: can we move it to `static defaultProps = {  id: generateId('checkbox'), }` ?

zelitzsi: No, because that will cause all components to share the same id. It needs to be an instance variable.

PR #231, opened 2018-10-23
fix field label spacing with help callout

zelitzsi:
## PR Checklist

* Field labels previously were taking up 21px when they had a help callout instead of the normal 11px. Now they will take up the right amount of space, since the help callout icon will not take up any space at all.


denysYefremov:
can you share screenshots how it was and how it becomes? Because it looks weird that you push it to render -15px top

zelitzsi:
@denysYefremov 
helptooltip

PR #239, opened 2018-11-28
SCLv6 canary 10 release fixes

zelitzsi:
## PR Checklist

* Fixed select animation
* Fixed navigation layout spacing
* Turned off spellcheck by default for inputs
* Fixed spacing on date pickers
* Fixed show password button in input
* Display the currently selected item value when select component is closed
* Fixed disabled state on links
* Locked down `react-popper` dependency
* Added `name` field to select's input component
* Changed dist file in `package.json`


a-type:
Did we want to do the rest props switch-up for Select here? Seems like we'd want to go ahead and make that decision before going gold.

zelitzsi:
@a-type I want to hold off on that for now. I think it's the right call, but I'd rather get this canary out and move onwards. Also, I'm planning on continuing with canary versions in the short term so that we can make sure the code here is stable.

PR #245, opened 2018-12-11
UI-648 Added horizontal alignment of content for Field

denysYefremov:
## PR Checklist

Added horizontal alignment for Field component


src/layouts/Field/Field.js
```diff
@@ -35,7 +35,10 @@ class Field extends React.Component {
      * (stretch|left|right): the alignment of contents within the Field
      */
     alignContent: PropTypes.oneOf(['stretch', 'left', 'right']),
-
+    /**
+     *  (top|center|bottom): the horizon alignment of content within the Field
```
zelitzsi: sp: horizontal

zelitzsi: And if it's `top|center|bottom`, wouldn't that be vertical alignment?

src/layouts/Field/Field.md
```diff
@@ -108,3 +108,38 @@ To make sure that forms are accessible, Fields supports associating labels with
   </Field>
 </Field.Group>
 ```
+
+### Providing a horizon alignment
```
zelitzsi: It's good that we have a place that is testing this code. However, I would prefer we leave testing to the Storybook pages and keep the documentation clean. You could use it in the toggle in the first example on this page.

src/layouts/Field/Field.js
```diff
@@ -125,15 +128,24 @@ class Field extends React.Component {
...
+    };
+
     return {
       gridArea: `${gridName}${column} / ${gridName}${column} / ${gridName}${endColumn} / ${gridName}${endColumn}`,
       // align-self determines how the parts vertically align if an adjacent part is taller.
       // labels should stick to the bottom if an adjacent label has two or more lines.
       // content and helpText should stick to the top if an adjacent element is taller than them.
-      alignSelf: gridName === 'label' ? 'end' : 'start',
+      alignSelf: alignContentHorizontally
+        ? horizontalAlignment[alignContentHorizontally]
```
zelitzsi: Maybe it would make sense to use "auto" alignment for the field contents in general?

src/layouts/Field/Field.js
denysYefremov: khm, I thought it calls `horizontal` OK will change it, thank you!

src/layouts/Field/Field.js
denysYefremov: we can do in that way, but then we won't know if the user has provided it, or it is a default value

zelitzsi: On second thought, it seems like we should default to the current behavior so we don't break anything. I know that `<textarea>` is a good example of something that shouldn't be centered with other fields.

PR #251, opened 2019-01-17
BWDB-4554 UI - Pre/Post Directional Fields Cannot be Cleared

safronenko23:
## PR Checklist

## Breaking Changes

Cleaning of selected item doesn't work if renderOption is specified. Because option will be null if you clear selection.

### DIFF
--- a/src/components/Select/Select.js
+++ b/src/components/Select/Select.js
@@ -154,21 +154,13 @@ class Select extends React.PureComponent {
         selectedItem={value}
         {...rest}
       >
-        {({ inputValue, ...downshiftProps }) => (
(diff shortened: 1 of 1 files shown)


src/components/Select/Select.js
```diff
@@ -154,21 +154,13 @@ class Select extends React.PureComponent {
         selectedItem={value}
         {...rest}
       >
-        {({ inputValue, ...downshiftProps }) => (
+        {({...downshiftProps}) => (
```
a-type: Can be simplified
```suggestion
        {downshiftProps => (
```

safronenko23: done

PR #254, opened 2019-01-24
Fix searchable select initial value

safronenko23:
## PR Checklist

## Breaking Changes

There was an issue with the initial value of select. When you provide any `value` prop to `searchable select` it provides ‘’ (empty string) as value for the select, so the only thing you could see was default placeholder. But if you opened the select box and close it without select anything, it selected the provided value. The reason is that we provided `inputValue: '',` by default for `searchable select` and update this value only on change event. My solution is set `inputValue`  to `undefined` because it will use value instead of `inputValue`.

### DIFF
--- a/src/components/Select/SearchableSelect.js
+++ b/src/components/Select/SearchableSelect.js
@@ -34,7 +34,7 @@ export default class SearchableSelect extends React.Component {
 
   state = {
     filteredOptions: this.props.options,
-    inputValue: '',
+    inputValue: undefined,
   };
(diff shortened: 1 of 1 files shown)


safronenko23:
> Were you able to test this with Janus in a local environment? I want to make sure it doesn't break our navigation.

Yes, I made tests locally:
1. Manual testing on Dashboard
2. Ran cypress tests (now passed)
3. Configured md file with this scenario.

PR #255, opened 2019-01-30
Fixed issues with displaying values by select component

denysYefremov:
## PR Checklist

The main issue was with the ability of `downshift` to re-run it's `itemToString` function when the data in this function loaded async. When select showed the value this value frizzed and didn't update when the data in `itemToString` was updated and this function returns new value. 

The idea of fix is to check that rendered data (`inputValue`) is not the same as the one which `itemToString` returns on component update. Do nothing if data the same and update it if the `itemToString` returns updated value.

Added `keepProvidedValue` flag to the `Select` component which allow you to show the value which user provided on clean select element instead of clean it completely (false by default)

The issues with previous solutions:

1. 
```
{({ inputValue, ...downshiftProps }) => (
    ...
    <CurrentValue
        {...downshiftProps}
        inputValue={
            downshiftProps.isOpen
                ? inputValue
                : inputValue ||
                  downshiftProps.itemToString(
                      downshiftProps.selectedItem,
                  )
        }
        ....
    />
)}
```

selectboxwrongbehaviour

2. 
```
{({ inputValue, ...downshiftProps }) => (
    ...
    <CurrentValue
        {...downshiftProps}
        inputValue={
            downshiftProps.isOpen
                ? inputValue
                : inputValue ||
                  downshiftProps.itemToString(
                      downshiftProps.selectedItem,
                  ) || ''
        }
        ....
    />
)}
```

selectboxwrongbehaviour2

The final version(current fix): 

selectboxbehaviour

all examples were tested in the styleguide by the next code:

```
class Test extends React.Component {
  constructor(props) {
   super(props);
    this.state = {
      options: ['1', '2', '3'],
      strings: { 1: 'One', 2 : 'Two', 3: 'Three'}
    }
  };
  
  componentDidMount() {
    const self = this;
    setTimeout(() => {
      self.setState(
        { 
          strings: { 1: 'One1', 2: 'Two1', 3: 'Three1'} 
        })
    }, 3000)
  }
  
  render() {
    return (
      <Select.Searchable
          options={this.state.options}
          renderOption={option => this.state.strings[option]}
          {...this.props}
       />
    )
  }
}

<div>
    <Test />
    <Test value={1} />
</div>;
```

### DIFF
--- a/src/components/Select/Select.js
+++ b/src/components/Select/Select.js
@@ -145,6 +145,10 @@ class Select extends React.PureComponent {
     } = this.props;
     const combinedLoading = loading || isLoading;
     let combinedPlaceholder = noneText !== undefined ? noneText : placeholder;
+    // This variable was presented, because of we don't have the `onOpen()` hook
+    // The only way to get information about select was just opened is to use `stateReducer`
+    // but the problem with this is that it mutates `options` to 1 element, currently selected,
+    // so you won't get the real item index in the `options`
     let justOpened = false;
 
     return (
@@ -157,20 +161,31 @@ class Select extends React.PureComponent {
       >
         {downshiftProps => {
           const { isOpen, inputValue, itemToString, selectedItem } = downshiftProps;
+          /* if select is closed */
           if (!isOpen) {
+            /* update justOpened to false, to know when exactly it will change to true first time */
             justOpened = isOpen;
+            /* downshift doesn't handle the changes inside of the `itemToString` function, so if your data
+             * inside of the renderOption(itemToString) will update async you won't know about this
+             * because the returned value won't change until you will change the select value */
             if (
               inputValue
               && inputValue !== itemToString(selectedItem)
             ) {
+              // we checked the variable, but because it is the value we need to assign value
+              // to the object variable, which is a link, not the value
               downshiftProps.inputValue = itemToString(selectedItem);
             }
           } else {
+            // Check that select was closed before and ignore if it was opened
             if (!justOpened) {
+              // Show that this is first time it was opened
               justOpened = isOpen;
+              // change the highlighted index to element which was selected before
               downshiftProps.setHighlightedIndex(options.indexOf(selectedItem));
             }
 
+            // Show previous value as a placeholder
             combinedPlaceholder = itemToString(selectedItem);
           }
           return ((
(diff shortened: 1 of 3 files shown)


a-type:
I'm not sure this behavior is correct. If the client has provided a value to Select and the user clicks the X button, it should reset that value to `null` by invoking `onChange(null)`. And I believe that Downshift should already do that for us.

The only time I could see the Select not working that way would be if it was `required`, in which case the X button should not show up at all.

I think we need UX involved to truly clarify this behavior.

a-type:
I still don't like the behavior of the second Select. When you click the X, the Select should always lose its value. But I think that it's only behaving that way because you have a hard-coded value in the demo instead of one that respects `onChange`.

src/components/Select/Select.js
```diff
@@ -146,33 +141,37 @@ class Select extends React.PureComponent {
       className,
       popperProps,
       name,
-      keepProvidedValue,
       ...rest
     } = this.props;
     const combinedLoading = loading || isLoading;
-    const combinedPlaceholder = noneText !== undefined ? noneText : placeholder;
+    let combinedPlaceholder = noneText !== undefined ? noneText : placeholder;
+    let justOpened = false;
```
a-type: Can you add comments explaining this behavior? I don't understand it, and I'm a little worried about adding complexity with side effects in render.

src/components/Select/Select.js
```diff
@@ -157,20 +161,31 @@ class Select extends React.PureComponent {
       >
         {downshiftProps => {
           const { isOpen, inputValue, itemToString, selectedItem } = downshiftProps;
+          /* if select is closed */
           if (!isOpen) {
+            /* update justOpened to false, to know when exactly it will change to true first time */
             justOpened = isOpen;
+            /* downshift doesn't handle the changes inside of the `itemToString` function, so if your data
```
a-type: I know that Downshift not knowing about changes to `itemToString` is part of the behavior you're trying to address, but I still feel like this is a very complicated solution, when we could just as easily put the Select in a `loading` state until all the data needed to render the items is loaded. It's actually a better user experience to not show the data until it is ready, to avoid the data suddenly changing in a confusing way. I'd prefer to take the simpler solution than introduce complex logic in this component for a particular use case and risk it breaking for others.

denysYefremov:
Just to warning you, now all the selects which are expecting that value will update dynamically won't work, previously at least @zelitzsi changes did the update

denysYefremov:
Wow, nice catch will fix this

denysYefremov:
Here is updated behaviour:

latestselectbehaviour2

PR #257, opened 2019-02-06
Add disabled state for file loader

a-type:
## PR Checklist

image

## Breaking Changes

* None

* FileLoader disabled state was not functional. I updated the Dropzone library to support the `disabled` prop and fixed the styling.

## New Visual Components

* None

## New Behavioral Components

* None

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14206,12 +14206,12 @@
       }
     },
     "react-dropzone": {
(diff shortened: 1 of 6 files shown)


zelitzsi:
@a-type Might be a good idea to fix the underline of `CLICK TO BROWSE` actually.

PR #258, opened 2019-02-08
Fix 'Invalid' state for DatePicker

safronenko23:
## PR Checklist

## Breaking Changes

Providing 'invalid=true' for DatePicker throws exception. Since get() deprecated and not imported. Changed it to var();

Fix 'Invalid' state for DatePicker.

### DIFF
--- a/src/components/DatePicker/styles/Wrapper.js
+++ b/src/components/DatePicker/styles/Wrapper.js
@@ -58,7 +58,7 @@ export default styled.div`
     css`
       .DateInput_input:not(.DateInput_input__focused) {
         border-color: var(--colors-negative-border) !important;
-        box-shadow: inset 0 -5px 0 var(--colors.negative.light')};
+        box-shadow: inset 0 -5px 0 var(--colors-negative-light)};
       }
     `};
 `;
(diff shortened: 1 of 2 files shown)


src/components/DatePicker/styles/Wrapper.js
```diff
@@ -1,4 +1,5 @@
 import styled, { css, keyframes } from 'styled-components';
+import get from 'extensions/themeGet';
```
denysYefremov: As I can see from the code, we mixed the `${get()}` and `var` here, I don't like that idea and think that we should choose one of the approaches. I'd prefer `var`

a-type:
@zelitzsi did we actually deprecate `get`? I thought it was required for IE11, but I don't remember where we came down on that. I'm good with dropping it, especially if we don't want to commit to providing a perfect experience in IE (everything will still be there, but the colors / sizings would be off if CSS vars don't work).

zelitzsi:
@a-type I left in support for `themeGet` primarily for compatibility. It would be nice to have a compatibility mode if it was easy to do, but otherwise I'm okay with using css vars.

safronenko23:
@a-type @zelitzsi let me know if I need to change something.

PR #261, opened 2019-02-14
BWDB-4711

antkoval:
## PR Checklist

## Breaking Changes

## New Visual Components

## New Behavioral Components

### DIFF
--- a/src/components/Button/SubmitButton.js
+++ b/src/components/Button/SubmitButton.js
@@ -71,7 +71,7 @@ class SubmitButton extends React.Component {
       return <span>{pristineContents}</span>;
     }
     if (loading) {
-      return <Loader size="16px" />;
+      return <Loader size="14px" />;
     }
 
     return children;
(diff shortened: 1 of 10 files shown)


src/components/Button/SubmitButton.js
```diff
@@ -71,7 +71,7 @@ class SubmitButton extends React.Component {
       return <span>{pristineContents}</span>;
     }
     if (loading) {
-      return <Loader size="14px" />;
```
tedwards-bw: On my display, setting Loader size to 16px does not keep the height constant between the states (Pristine, Loading, etc...).  Also, do we want the Ring size to increase, or do we want the overall button height to be constant with contents centered vertically? You might want to check w/ other Forge team members about the best way to fix it.

zelitzsi: @tedwards-bw The requirements should come from UX.

antkoval: changed the loader size back and adjusted buttons min-height

tedwards-bw: Should the min-height be confined to Button.Submit (the only thing that replaces its contents with a Loader)?  Would Button.Small, for example, be impacted by this?

antkoval: I changed the solution to affect only submit button

PR #263, opened 2019-02-28
Added css style for blocked state of day for DatePicker

safronenko23:
## PR Checklist

## Breaking Changes

Currently we use `isOutsideRange` property for DatePicker to prevent out of range dates. But in this case you cannot handle `out of range` date if you set it by hands (print date). Because `onDateChange` will have date as `null`.

I found a solution how to solve this. We can change `isOutsideRange` to `isDayBlocked` property.
In this case everything works fine. But we don’t have  ccs style for `Blocked` state of DatePicker.
Because currently `isDayBlocked` days looks as `Available for choose`

### DIFF
--- a/src/components/DatePicker/styles/overridesCss.js
+++ b/src/components/DatePicker/styles/overridesCss.js
@@ -177,7 +177,8 @@ export default css`
     color: ${get('colors.text.inverted')};
   }
 
-  .CalendarDay__blocked_out_of_range {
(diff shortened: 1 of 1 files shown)


safronenko23:
@a-type @zelitzsi May I ask you to publish this and build new version of SCL if everything is fine. Thanks

zelitzsi:
@safronenko23 We'll release it with #262.

PR #267, opened 2019-03-06
[BWDB-4731] Correct props

antkoval:
## PR Checklist

* DatePicker don't get `invalid` and `className` props passed (no such prop)
* FileLoader don't get `error` prop (no such prop)
* Inpur has no `aria-role` prop (changed to `role`)
* Options has different way of passing `ref` and `style` (`getMenuProps() is not a function` error)
* Radio can accept not only `string` (in some cases `bool` is required)

### DIFF
--- a/src/components/Radio/Radio.js
+++ b/src/components/Radio/Radio.js
@@ -19,7 +19,7 @@ class Radio extends React.PureComponent {
     /**
      * The value of the radio; a string that describes what the value should be if checked=true
      */
-    value: PropTypes.string.isRequired,
+    value: PropTypes.any.isRequired,
     /**
      * Whether the radio is checked
      */
--- a/src/components/Radio/Radio.js
+++ b/src/components/Radio/Radio.js
@@ -19,7 +19,7 @@ class Radio extends React.PureComponent {
     /**
      * The value of the radio; a string that describes what the value should be if checked=true
      */
-    value: PropTypes.any.isRequired,
+    value: PropTypes.string.isRequired,
     /**
      * Whether the radio is checked
      */
--- a/src/components/Select/styles/Options.js
+++ b/src/components/Select/styles/Options.js
@@ -29,7 +29,8 @@ const Options = forwardRef(
       return (
         <OptionsListContainer
           data-placement={placement}
-          {...getMenuProps({ ref, style })}
+          ref={ref}
+          style={style}
         >
           <OptionsList>
             <Option>
@@ -43,10 +44,8 @@ const Options = forwardRef(
     return (
       <OptionsListContainer
         data-placement={placement}
-        {...getMenuProps({
-          ref,
-          style,
-        })}
+        ref={ref}
+        style={style}
       >
         <OptionsList>
           {options.map((option, index) => (
--- a/src/components/Select/styles/Options.js
+++ b/src/components/Select/styles/Options.js
@@ -29,8 +29,7 @@ const Options = forwardRef(
       return (
         <OptionsListContainer
           data-placement={placement}
-          ref={ref}
-          style={style}
+          {...getMenuProps({ ref, style })}
         >
           <OptionsList>
             <Option>
@@ -44,8 +43,10 @@ const Options = forwardRef(
     return (
       <OptionsListContainer
         data-placement={placement}
-        ref={ref}
-        style={style}
+        {...getMenuProps({
+          ref,
+          style
+        })}
       >
         <OptionsList>
           {options.map((option, index) => (
--- a/src/components/Step/Step.js
+++ b/src/components/Step/Step.js
@@ -61,7 +61,9 @@ export default class Step extends React.Component {
     return (
       <Container {...rest}>
         <Number complete={complete} number={number} />
-        <Content>{children}</Content>
+        {children ?
+          <Content>{children}</Content> : null
+        }
       </Container>
     );
   }
--- a/src/components/Step/Step.js
+++ b/src/components/Step/Step.js
@@ -14,7 +14,7 @@ export default class Step extends React.Component {
     /**
      * Content of the step
      */
-    children: PropTypes.node.isRequired,
+    children: PropTypes.node,
     /**
      * Step number
      */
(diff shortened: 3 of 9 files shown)


src/layouts/SplitContentLayout/SecondaryContent/styles/SidebarList/Item.js
```diff
@@ -27,7 +27,7 @@ export default class SidebarListItem extends React.PureComponent {
     /**
      * A component for rendering a container of an item
      */
-    Container: PropTypes.func,
+    Container: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
```
denysYefremov: I think it should be `PropTypes.oneOfType([PropTypes.Node, PropTypes.func])`

src/layouts/SplitContentLayout/SecondaryContent/styles/SidebarList/index.js
```diff
@@ -30,7 +30,7 @@ export default class SidebarList extends React.PureComponent {
     /**
      * A compoennt to render the container around the list
      */
-    Container: PropTypes.func,
+    Container: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
```
denysYefremov: I think it should be `PropTypes.oneOfType([PropTypes.object, PropTypes.func])`

stories/DatePicker.js
```diff
@@ -1,19 +1,29 @@
-import React from 'react';
+import React, { PropTypes } from 'react';
```
denysYefremov: It is deprecated import of `PropTypes`, you should take them from `import PropsTypes from 'prop-types';`

src/components/Step/Step.js
```diff
@@ -14,7 +14,7 @@ export default class Step extends React.Component {
     /**
      * Content of the step
      */
-    children: PropTypes.node.isRequired,
+    children: PropTypes.node,
```
denysYefremov: if you can avoid passing this prop, than you should have or
1. Default children
2. Skip render `<Content />`

zelitzsi:
@antkoval Can you provide some comments in Github on your code describing the errors we are fixing?

src/layouts/SplitContentLayout/SecondaryContent/styles/SidebarList/Item.js
a-type: These are Component classes being passed in, not rendered nodes. That's why `func` was used originally.

src/components/Radio/Radio.js
```diff
@@ -19,7 +19,7 @@ class Radio extends React.PureComponent {
     /**
      * The value of the radio; a string that describes what the value should be if checked=true
      */
-    value: PropTypes.string.isRequired,
+    value: PropTypes.any.isRequired,
```
a-type: What use case would not be `string` here? This prop is passed through to the `<input>` element, which only supports primitive attributes for value.

src/components/Select/styles/Options.js
```diff
@@ -29,7 +29,8 @@ const Options = forwardRef(
       return (
         <OptionsListContainer
           data-placement={placement}
-          {...getMenuProps({ ref, style })}
+          ref={ref}
```
a-type: Why are you removing Downshift supplied properties? These are a core part of how that library works.

a-type:
Please fill out the issue template, it's not clear what problems you're trying to solve here.

src/components/Radio/Radio.js
antkoval: There is a case where we need to pass a `bool` value, which is forbidden here

src/components/Select/styles/Options.js
antkoval: In this particular case `getMenuProps` caused an error `getMenuProps() is not a function`

antkoval:
The following PR is a resolution for multiple errors in portin pipeline. In many cases, props were passed to components in a wrong way (like passing className or invalid to DatePicker, where it doesn't exist) or behavior of components allow null values for required fields. Also `area-role` is not supported by React. Most of portin errors are fixed on janus side (https://github.com/Bandwidth/janus/pull/767), but some are not solvable only in janus

src/components/Step/Step.js
antkoval: skipped rendering content

src/layouts/SplitContentLayout/SecondaryContent/styles/SidebarList/Item.js
denysYefremov: // A React element.
  optionalElement: PropTypes.element,

then `element` should work

src/components/Radio/Radio.js
zelitzsi: @antkoval A boolean value is not a valid option for the `input` tag's `value` attribute. https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement

src/components/Select/styles/Options.js
zelitzsi: @antkoval In that case, we should do something more like:
```
const menuProps = getMenuProps ? getMenuProps({ref, style}) : {};
```

and then spread `menuProps`.

antkoval: @zelitzsi  the problem here is not about wrong `getMenuProps()` function, but in broken mechanism of its spreading. The fix you offer turns Options in empty input and behaves incorrectly

zelitzsi: Why is it spreading incorrectly? As @a-type said, I think `getMenuProps()` is required for this input to receive everything it needs to from Downshift.

antkoval: @zelitzsi `ref` and `style` fields are present in props, but they can not be spread using the existing code. IDK why

src/components/Radio/Radio.js
a-type: So the use case is that you have two radio buttons where one is "true" and one is "false"? That seems like a good scenario for a Toggle instead.

If you really need it, convert the values to strings. We don't want to supply invalid attributes to html elements.

src/components/Select/styles/Options.js
a-type: It's especially important to send `ref` through `getMenuProps` since Downshift applies its own ref to the element and it must internally combine the refs.

src/components/Radio/Radio.js
antkoval: revert these changes then

src/components/Select/styles/Options.js
antkoval: this component works incorrectly on localhost, thou on env it's ok. Revert these changes, but it's still unclear for me why this can happen

PR #271, opened 2019-03-25
fix file loader text link preventing click

a-type:
GIF

## PR Checklist

* Switched Link component in FileLoader to use the TextLink styling component instead to avoid Link behaviors.


a-type:
I think we'd need Cypress to actually test this effectively since it invokes native behavior.

Also, we didn't merge the test PR yet.

PR #274, opened 2019-04-03
Bringing Modals into alignment with UX rules

braxtonCoats:
## PR Checklist

## Breaking Changes

1. removed title callout that would appear on hover on modal elements
2. removed padding on action content to extend seperation line to full length of modal.
3. Added 30px padding to button container to preserve correct spacing.
4. Added border rounding to title-top-left and title-top-right.

## New Visual Components

## New Behavioral Components


src/components/Modal/Modal.js
```diff
@@ -174,6 +174,7 @@ class Modal extends React.Component {
           id={id}
           className={className}
           {...rest}
+          title=""
```
a-type: This looks wrong. Maybe we can just remove the `title` prop from `Window`.

zelitzsi:
@braxtonCoats Can you include a screenshot that shows what it looks like now?

braxtonCoats:
added a screen of what it looks like. Note the ```title=""``` removes a callout that appears when a modal element (say button) is hovered.

src/components/Modal/Modal.js
braxtonCoats: Will try

PR #275, opened 2019-04-05
Alternative accordions

braxtonCoats:
## PR Checklist

~- [ ] Props have JSDoc comments~
~- [ ] Any extra props are spread into the outermost rendered element~

## Breaking Changes

Changed default text color from bandwidth blue to normal text color (#666666). Fixed spacing issues. Added an alternative blue versions for specific use cases. 

Screen Shot 2019-04-05 at 1 49 58 PM

## New Visual Components

## New Behavioral Components


src/components/Accordion/styles/LabelText.js
```diff
@@ -3,5 +3,6 @@ import styled from 'styled-components';
 // FIXME: it would be nice to craft the right CSS so that this element is unecessary.
 export default styled.span`
   margin: auto;
+  padding-bottom: 1px;
```
mike-zarandona: If you have to do hacky stuff like a `1px` `padding-bottom`, that may mean that the `display` property isn't set correctly on this element.  Possibly try `display: inline-block` to see if that fixes the layout of the box model versus forcing it with padding.

mike-zarandona: After a F2F conversation, this comment may not be 100% relevant.  Dismiss if that's the case  👍

src/components/Accordion/styles/Label.js
```diff
@@ -3,8 +3,8 @@ import get from 'extensions/themeGet';
 import userSpacing from 'extensions/userSpacing';
 
 const AccordionLabel = styled.div`
-  padding: ${userSpacing.text};
-  color: ${get('colors.primary.default')};
+  padding: 30px;
```
a-type: `userSpacing.text` actually takes into account the cap-height of the font to provide an accurate spacing from the real text. I don't think it should be removed, it probably just needs to be configured.

a-type:
Should #274 be closed then?

a-type:
Ok, should this one be closed?

PR #278, opened 2019-04-10
Added rounding to shadow container and removed -1px.

braxtonCoats:
## PR Checklist

## Breaking Changes

Removed styling that points out its used for browser quirks. While working on these changes I didn't notice anything odd around the shadows rendering

Added border rounding and extended heading and separating lines to reach the end of the table.

Screen Shot 2019-04-10 at 3 34 24 PM


a-type:
The reason that hack was there was so that the right internal shadow would show up. Can you verify that the right internal shadow is present when there is more table content beyond the right border?

braxtonCoats:
Screen Shot 2019-04-18 at 11 11 05 AM
 
Table shadow still shows when there is content outside of the visible area

a-type:
Looks good, but which browser are you using? Have you tried others?

braxtonCoats:
Yea, just tested edge, IE and firefox (mac and windows) and all look good.

PR #279, opened 2019-04-17
Kill autofill via new Input prop

mike-zarandona:
## PR Checklist

## Breaking Changes

* No breaking changes with this update

* ~The `Input` component now has a new property called `killAutofill`.~
* The `Input` component's `autoComplete` property has been enhanced to implement a `readonly` property until `focus`ed

### DIFF
--- a/src/components/Input/Input.js
+++ b/src/components/Input/Input.js
@@ -132,6 +132,11 @@ class Input extends React.PureComponent {
      * Hides the cursor from view
      */
     hideCursor: PropTypes.bool,
+    /**
+     * Actively blocks autofill/password managers by applying a `readonly`
+     * propery until the field is `focus`ed.
+     */
+    killAutofill: PropTypes.oneOfType([PropTypes.bool, PropTypes.string]),
   };
 
   static defaultProps = {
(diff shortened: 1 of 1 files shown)


src/components/Input/Input.js
```diff
@@ -277,6 +294,7 @@ class Input extends React.PureComponent {
         type={type}
         onBlur={this.onBlur}
         autoComplete={this.getAutoComplete()}
+        killAutofill={killAutofill}
```
zelitzsi: This prop is unneeded.

src/components/Input/Input.js
```diff
@@ -132,6 +132,11 @@ class Input extends React.PureComponent {
      * Hides the cursor from view
      */
     hideCursor: PropTypes.bool,
+    /**
+     * Actively blocks autofill/password managers by applying a `readonly`
+     * propery until the field is `focus`ed.
+     */
+    killAutofill: PropTypes.oneOfType([PropTypes.bool, PropTypes.string]),
```
zelitzsi: This shouldn't include `string` because it doesn't support them.

PR #280, opened 2019-04-19
BWDB 4856 - update remote link so icon isn't obscured by subsequent text

pisaychuk:
## PR Checklist

If there is a child Link element pointing an external resource it has new tab icon fading in on hover.
This PR changes make that icon to overlap the element which goes close to the Link, for example text with a circled background white by default or derived from an Alert if out if put in Alert

### DIFF
--- a/src/components/Alert/styles/Border.js
+++ b/src/components/Alert/styles/Border.js
@@ -1,6 +1,7 @@
 import styled, { keyframes } from 'styled-components';
 import get from 'extensions/themeGet';
 import userSpacing from 'extensions/userSpacing';
+import TextLink from 'components/Link/styles/TextLink';
 
 const fadeIn = keyframes`
   from { opacity: 0; }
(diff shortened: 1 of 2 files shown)


src/components/Alert/styles/Border.js
```diff
@@ -19,17 +30,10 @@ const AlertBorder = styled.div`
...
-      case 'error':
-        return get('colors.negative.light');
-      default:
-        return get('colors.primary.light');
-    }
-  }};
-
+  background: ${getBgColorForType};
+  
+  & a::before { background-color: ${getBgColorForType} !important;}
```
a-type: Is `!important` needed here? Can you just make this selector more specific than the default color selector ([link](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity))? For instance, you can add additional `&` to increase specificity sometimes. You can also import `TextLink` here and interpolate it to access its class name: `& ${TextLink}::before { ... }`.

PR #285, opened 2019-05-14
Fixing vulnerabilities with dependencies

tbowenbw:
Fixing security vulnerabilities.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "@storybook/addon-links": "3.4.11",
     "@storybook/react": "4.0.0-rc.2",
     "babel-loader": "8.0.4",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -92,7 +92,7 @@
     "@storybook/addon-links": "3.4.11",
     "@storybook/react": "4.0.0-rc.2",
     "babel-loader": "8.0.4",
-    "babel-plugin-inline-react-svg": "1.0.0",
+    "babel-plugin-inline-react-svg": "^1.1.0",
```
zelitzsi: Can we use an exact version for this?

tbowenbw: sure, this was just the result of running the command that the audit suggested

PR #287, opened 2019-05-21
Updated icons

braxtonCoats:
## PR Checklist

- ~[ ] Visual and behavioral components are separated~
- ~[ ] Props have JSDoc comments~
- ~[ ] All relevant visual sub-components can be overridden via props~
- ~[ ] Any extra props are spread into the outermost rendered element~

Updated Icons to conform to icons that UX uses. This significantly shortened the list of used icons.

Added support for legacy icons that weren't originally included in the Bandwidth_2019 font file.

Screen Shot 2019-05-23 at 11 31 03 AM


src/components/Icon/icons.js
```diff
@@ -1,229 +1,41 @@
 export const map = {
```
zelitzsi: Are we positive this is the full set of icons? It would be nice to test this against Janus to make sure we aren't removing icons are that being used...

src/components/Icon/icons.js
```diff
@@ -1,229 +1,41 @@
...
+  copy: '\\f109',
+  csv: '\\f10a',
+  delete: '\\f10b',
+  download: '\\f10c',
+  edit: '\\f10d',
+  envelope: '\\f10e',
+  expired: '\\f10f',
+  fileFilled: '\\f110',
+  file: '\\f111',
+  filterClear: '\\f112',
```
ssodano: Looks like clearFilters was renamed to filterClear, won't this break backwards compatibility?

a-type: Yup. @braxtonCoats try to avoid changing the name of icons that bear the same general meaning as the previous set.

braxtonCoats: Yes, good catch. I'll adjust that. Also many icons are being removed here to clean up things. I'm not 100% sure if any of the icons removed were used anywhere else.

ssodano:
The following appear to be used still:

- back
- clearShoppingCart2
- delete2
- delete3
- delete_3
- plusMath
- trash
- viewMore

braxtonCoats:
updated to included those icons @ssodano @zelitzsi @mike-zarandona

a-type:
Here's a list of icons used in Janus:

- search
- attention
- help
- plusMath
- edit
- checkmark
- ok
- down
- expandArrow
- edit2
- file
- fileFilled
- businesswoman
- assistant

braxtonCoats:
@a-type All icons listed before have been accounted for now.

PR #296, opened 2019-06-24
disable autocomplete always for select components

zelitzsi:
## PR Checklist

* Autocomplete is automatically disabled for all Select components. I can't think of a reason that Select should ever have autocomplete.


src/components/Select/styles/CurrentValue.js
```diff
@@ -73,6 +73,7 @@ const Unsearchable = forwardRef(
           value={inputValue}
           hideCursor
           {...rest}
+          autoComplete={false}
```
a-type: As noted from experience, `autoComplete={false}` doesn't necessarily work. That's part of the reason these components were changed to allow user-defined autoComplete.

Maybe we could just add a semi-random string here instead.

zelitzsi: I tested against Janus with a random string and it still autocompleted. That said, autoComplete=off seems to be respected now by at least my browser, so I changed that to be the default for `<Select>` (while still allowing devs to override the value).

PR #299, opened 2019-07-12
simplify field implementation

a-type:
The objective of this PR is to replace the existing Field implementation with a far simpler one which addresses all the bugs in it today, without changing the fundamental usage so as to require little to no migration.

There are some changes to how field lays out now:

1. HelpText is no longer pushed to the bottom of the available content area. It sits below the actual content of the field (this is good ✔️ , the old behavior was strange and lead to helptext being visually separate from its subject content)
2. Labels are now always present; if the user doesn't specify a `label` prop, a blank label is added which takes up the height of one label line.
	a. This introduces a 'soft requirement' that labels should always be 1 line high to align properly. I have never seen this requirement violated though.
3. DOM structure is simplified; rows no longer exist as elements in the DOM (or even as an abstraction whatsoever).
4. Field content must now be at least 53px tall. This makes things like checkboxes and toggles align with inputs seamlessly.
4. Class names are added to field constituent elements


a-type:
The main field example from the docs: 

image


Column span example: 

image


Customization example:

image

a-type:
Looks like I missed help callouts.

a-type:
Updated example:

image

PR #300, opened 2019-07-22
added onInputValueChange prop to SearchableSelect

SmithPR:
## PR Checklist

- [✔️] Visual and behavioral components are separated
- [❔] Exported components are documented with examples
- [✔️] Props have JSDoc comments
- [✔️] All relevant visual sub-components can be overridden via props
- [✔️] Any extra props are spread into the outermost rendered element

## Breaking Changes

N/A

Select.Searchable can now accept an onChange callback, `onInputValueChange`.
## New Visual Components

N/A

## New Behavioral Components

N/A


src/components/Select/SearchableSelect.js
```diff
@@ -59,6 +64,8 @@ export default class SearchableSelect extends React.Component {
     this.setState({
       inputValue: inputValue,
       filteredOptions: filterOptions(options, inputValue, this.props),
+    }, ()=>{
+      props.onInputValueChange(inputValue);
```
a-type: Does this need to happen after state is updated?

Also, please add a quick existence check a la `props.onInputValueChange && props.onInputValueChange(inputValue);`

SmithPR: I think so: in case the user wants to update the options, we do not want them to then be overwritten by the asynchronous setState, as unlikely as that may be.

👍 for the existence check

a-type:
RTM

PR #303, opened 2019-08-08
Use default classes

zelitzsi:

* Components have default classnames to make selecting them easier
* Updated the React peer dependency to be accurate


src/components/Note/Note.js
```diff
@@ -4,7 +4,7 @@ import * as styles from './styles';
 import { defaultProps } from 'recompose';
 
 const Note = ({ children, Container, Corner, ...rest }) => (
-  <Container {...rest}>
+  <Container  {...rest}>
```
mike-zarandona: Is this a Prettier thing?

PR #309, opened 2019-09-18
fix nav spacing

zlitzsinger-bw:

* The navigation component was having specificity issues in styled-components 4.2.0, so I removed the offending CSS. I believe this CSS didn't actually do anything useful anyway.


a-type:
It's for stacked item lists. That particular rule would match if we had two full-size lists on top of one another, which never happens.

vinks:
ср, 18 сент. 2019 г., 6:44 ПП Grant Forrest <notifications@github.com>:

